PR #129, opened 2024-10-21
feat: Extract L1 gas cost parameters from deposit transaction

### LINKED ISSUE #125: Extract base fee parameters for L1 cost from blocks
The implemented L1 cost algorithm depends on parameters extracted from blocks.

These parameters are different per block and are encoded in the first transaction that is supposed to be the deposit type.

RomanHodulak:
Closes #125

### DIFF
--- a/moved/src/move_execution/gas.rs
+++ b/moved/src/move_execution/gas.rs
@@ -121,12 +121,45 @@ impl L1GasFee for EcotoneL1GasFee {
     }
 }
 
+/// Creates algorithm for calculating cost of publishing a transaction to layer-1 blockchain.
+pub trait CreateL1GasFee {
+    /// Extracts parameters from deposit transaction and creates the algorithm for calculating L1
+    /// gas cost.
+    fn for_deposit(&self, data: &[u8]) -> impl L1GasFee + '_;
+}
+
+pub struct CreateEcotoneL1GasFee;
+
+impl CreateL1GasFee for CreateEcotoneL1GasFee {
+    fn for_deposit(&self, data: &[u8]) -> impl L1GasFee + '_ {
+        let l1_base_fee = U256::from_be_slice(&data[36..68]);
+        let l1_blob_base_fee = U256::from_be_slice(&data[68..100]);
+        let l1_base_fee_scalar =
+            u32::from_be_bytes(data[4..8].try_into().expect("Slice should be 4 bytes"));
+        let l1_blob_base_fee_scalar =
+            u32::from_be_bytes(data[8..12].try_into().expect("Slice should be 4 bytes"));
+
+        EcotoneL1GasFee::new(
+            l1_base_fee,
+            l1_base_fee_scalar,
+            l1_blob_base_fee,
+            l1_blob_base_fee_scalar,
+        )
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
 
     impl L1GasFee for U256 {
         fn l1_fee(&self, _input: L1GasFeeInput) -> U256 {
+            *self
+        }
+    }
+
+    impl CreateL1GasFee for U256 {
+        fn for_deposit(&self, _data: &[u8]) -> impl L1GasFee + '_ {
             *self
         }
     }
--- a/moved/src/state_actor.rs
+++ b/moved/src/state_actor.rs
@@ -1,4 +1,3 @@
-use alloy_rlp::Encodable;
 pub use payload::{NewPayloadId, NewPayloadIdInput, StatePayloadId};
 
 use {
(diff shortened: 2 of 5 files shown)


moved/src/state_actor.rs
```diff
@@ -225,11 +225,19 @@ impl<
...
         let mut total_tip = U256::ZERO;
         let mut outcomes = Vec::new();
+        let first_tx = transactions.next();
+        let l1_fee = match first_tx.as_ref() {
+            Some((_, NormalizedExtendedTxEnvelope::DepositedTx(deposit_tx), _)) => {
+                Some(self.l1_fee.for_deposit(deposit_tx.data.as_ref()))
+            }
+            _ => None,
+        };
+        let transactions = iter::once(first_tx).flatten().chain(transactions);
```
bulenttastan, line 240: nit: This line rebuilds the `transactions` I believe. Maybe we should leave a comment.

moved/src/state_actor.rs
```diff
@@ -225,11 +225,19 @@ impl<
 
     fn execute_transactions(
         &mut self,
-        transactions: impl Iterator<Item = (B256, NormalizedExtendedTxEnvelope, L1GasFeeInput)>,
+        mut transactions: impl Iterator<Item = (B256, NormalizedExtendedTxEnvelope, L1GasFeeInput)>,
         base_fee: U256,
     ) -> ExecutionOutcome {
         let mut total_tip = U256::ZERO;
         let mut outcomes = Vec::new();
+        let first_tx = transactions.next();
```
bulenttastan, line 233: nit: I'd leave that comment where the first transaction is always a `DepositTx`.

bulenttastan (approved):
Looks good to me ✈️

moved/src/move_execution/gas.rs
```diff
@@ -121,6 +121,33 @@ impl L1GasFee for EcotoneL1GasFee {
     }
 }
 
+/// Creates algorithm for calculating cost of publishing a transaction to layer-1 blockchain.
+pub trait CreateL1GasFee {
+    /// Extracts parameters from deposit transaction and creates the algorithm for calculating L1
+    /// gas cost.
+    fn for_deposit<'f, 'r>(&'f self, data: &'f [u8]) -> impl L1GasFee + 'r;
```
birchmd, line 128: I'm curious why we need the second lifetime parameter here. Seems like all the implementations have it resolve to `'static` and I can't think of a situation where we would have something other than that.

moved/src/state_actor.rs
birchmd, line 240: I think a better way to do this is to make the original iterator [peekable](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.peekable) and [peek](https://doc.rust-lang.org/std/iter/struct.Peekable.html#method.peek) at the first element instead of popping it off and then making a new iterator from this chained with the tail.

moved/src/state_actor.rs
RomanHodulak: I don't know the spec to reference if there is one as I got it from op-geth implementation.

moved/src/move_execution/gas.rs
RomanHodulak: I was able to use implicit lifetime and anonymouse lifetime combo for less verbose syntax.

Without the second lifetime we get the error:

```
error[E0505]: cannot move out of `transactions` because it is borrowed
   --> moved/src/state_actor.rs:242:45
    |
232 |         let mut transactions = transactions.peekable();
    |             ---------------- binding `transactions` declared here
233 |         let first_tx = transactions.peek();
    |                        ------------ borrow of `transactions` occurs here
...
242 |         for (tx_hash, tx, l1_cost_input) in transactions {
    |                                             ^^^^^^^^^^^^ move out of `transactions` occurs here
...
248 |                 l1_fee
    |                 ------ borrow later used here

```

I think this was the reason why I threw away the `peekable` which I originally used, glad you reminded me to put it back.

bulenttastan: This is the only spec I see that refers to first transaction being a deposit txn.
https://github.com/ethereum-optimism/optimistic-specs/blob/main/specs/deposits.md#kinds-of-deposited-transactions

We can just simply mention that based on the op-geth implementation the first transaction must be a deposit transaction.

Other than that I think this PR is good to merge ⛙

moved/src/state_actor.rs
bulenttastan: This is the only spec I see that refers to first transaction being a deposit txn.
https://github.com/ethereum-optimism/optimistic-specs/blob/main/specs/deposits.md#kinds-of-deposited-transactions

We can just simply mention that based on the op-geth implementation the first transaction must be a deposit transaction.

Other than that I think this PR is good to merge

RomanHodulak: Great find! Also, that repository is archived and deprecated, the specs have moved to https://github.com/ethereum-optimism/specs/

birchmd (approved):
Great job!

PR #130, opened 2024-10-21
feat: Charge L1 cost to transaction signer 

### LINKED ISSUE #126: Charge L1 cost to accounts in the blockchain state
The calculated L1 cost has to be charged during move execution along with gas_limit * max_fee.

It is not refunded at the end of the execution.

The changes need to be reflected in the Moved blockchain state.

RomanHodulak:
Closes #126

### DIFF
--- a/moved/src/move_execution/tests.rs
+++ b/moved/src/move_execution/tests.rs
@@ -69,7 +69,8 @@ fn test_execute_counter_contract() {
         bcs::to_bytes(&entry_fn).unwrap(),
     );
 
-    let outcome = execute_transaction(&tx, &tx_hash, state.resolver(), &genesis_config, 0).unwrap();
+    let outcome =
+        execute_transaction(&tx, &tx_hash, state.resolver(), &genesis_config, 0, &()).unwrap();
     state.apply(outcome.changes).unwrap();
 
     // Calling the function with an incorrect signer causes an error
@@ -88,7 +89,8 @@ fn test_execute_counter_contract() {
         TxKind::Call(EVM_ADDRESS),
         bcs::to_bytes(&entry_fn).unwrap(),
     );
-    let err = execute_transaction(&tx, &tx_hash, state.resolver(), &genesis_config, 0).unwrap_err();
+    let err =
+        execute_transaction(&tx, &tx_hash, state.resolver(), &genesis_config, 0, &()).unwrap_err();
     assert_eq!(
         err.to_string(),
         "Signer does not match transaction signature"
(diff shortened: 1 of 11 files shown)


server/src/lib.rs
```diff
@@ -55,6 +56,7 @@ struct Claims {
 
 const EIP1559_ELASTICITY_MULTIPLIER: u64 = 6;
 const EIP1559_BASE_FEE_MAX_CHANGE_DENOMINATOR: U256 = U256::from_limbs([250, 0, 0, 0]);
+const TREASURY: AccountAddress = AccountAddress::ZERO; // todo what is the real address?
```
bulenttastan, line 59: I believe the treasury address is: `0x4200000000000000000000000000000000000011` based on https://docs.optimism.io/stack/transactions/fees#fee-collection-and-distribution

birchmd: +1 for making it configurable.

birchmd (commented):
Looks pretty good! Please add a new test where the L1 fee is actually charged to an account so that we can see it working.

bulenttastan (dismissed):
Additional tests look good.

moved/src/move_execution/tests.rs
```diff
@@ -484,6 +515,107 @@ fn test_eoa_base_token_transfer() {
...
+    let sender_balance = quick_get_eth_balance(&sender.to_move_address(), state.resolver());
+    let receiver_balance = quick_get_eth_balance(&receiver.to_move_address(), state.resolver());
+    let treasury_balance = quick_get_eth_balance(&eth_treasury, state.resolver());
+    assert_eq!(sender_balance, mint_amount - transfer_amount - l1_cost);
+    assert_eq!(receiver_balance, transfer_amount);
+    assert_eq!(treasury_balance, l1_cost);
+}
+
+#[test]
+fn test_treasury_does_not_charge_l1_cost_on_failure() {
```
birchmd, line 569: This actually isn't the right behavior. If a transaction fails then the user must still be charged for the gas and L1 cost because we still did the work to process the transaction and it will still be persisted on the L1. The only time we do not charge is when the transaction is invalid such that it was not executed at all in the first place (e.g. wrong signature, wrong nonce, not enough balance to cover gas/L1 fees, etc).

bulenttastan: My understanding was that if a transaction that the L2 receives fails, it won't be sent to L1 at all. I guess failed transactions are persisted on L1 too.

birchmd (approved):
Nice work!

PR #132, opened 2024-10-24
Chore(deps): Update rust and crate versions

birchmd:
### Description
Updates Rust to the latest version. This is needed because the EVM library I want to use for the EVM native ([revm](https://crates.io/crates/revm)) generally requires the latest version of Rust.

While I was at it I decided to also run `cargo update` to get on the latest supported version of the various crates we depend on.

### Changes
- Some doc changes we needed to make clippy happy.

### Testing
Existing tests


Cargo.lock
```diff
@@ -292,14 +295,15 @@ dependencies = [
...
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3223d71dc78f464b2743418d0be8b5c894313e272105a6206ad5e867d67b3ce2"
+checksum = "6e701fc87ef9a3139154b0b4ccb935b565d27ffd9de020fe541bf2dec5ae4ede"
 dependencies = [
  "alloy-consensus",
  "alloy-eips",
  "alloy-json-rpc",
- "alloy-primitives",
+ "alloy-network-primitives",
+ "alloy-primitives 0.7.7",
```
RomanHodulak, line 306: Looks like we have multiple alloy versions now

RomanHodulak (approved):
The changes in rust lang seem pretty harmless.

:cool:

bulenttastan (approved):

Cargo.lock
birchmd: Good catch. I fixed it by upgrading to the latest version of the alloy library; which is probably good for us to do anyway. The downside is now we depend on multiple versions of `derive_more`, but unless we change our Aptos dependency then we can't fix that. I don't really want to change which Aptos we depend on because I think newer versions have some breaking changes for us.

https://github.com/MovedNetwork/op-move/pull/132/commits/75eca1b9270f8451c8c9ee3095167595adc21860

PR #136, opened 2024-10-30
refactor: Introduce `engine-api` crate

RomanHodulak:
Separates parts responsible for request and response serialization, routes definition and request handling into a dedicated crate for Ethereum JSON-RPC.


moved/src/block/gas.rs
```diff
@@ -101,26 +101,31 @@ impl GasFee for Eip1559GasFee {
     }
 }
 
-#[cfg(test)]
-mod tests {
+#[cfg(any(feature = "test-doubles", test))]
```
bulenttastan, line 104: What do we mean by `test-doubles` here? Test duplicate or a second test?

RomanHodulak: Inspired by https://martinfowler.com/bliki/TestDouble.html

bulenttastan (commented):
I think the `test-doubles` erases the boundary between test and production code as I see it being in the default features. I'm not so sure about this, so I'll let @birchmd decide.
Other than that I think the crate separation looks good.

moved/src/block/gas.rs
bulenttastan: After talking about the purpose of test doubles, realized that test only code doesn't carry between crates, so it's needed for the tests from external crates.

bulenttastan (dismissed):
Looks good. Let's create this new crate.

moved/src/types/mod.rs
```diff
@@ -1,7 +1,4 @@
 pub mod engine_api;
```
bulenttastan, line 1: Why don't we move this type into `engine-api` crate?

bulenttastan: Or alternatively we might want to separate the `moved/types` folder into its own crate named `types`, because that's where the dependency connection happens. This way `engine-api` will not depend on `moved` crate and vice versa. We will need to move `moved/primitives.rs` into the `types` crate too.

Cargo.toml
```diff
@@ -5,12 +5,14 @@ members = [
     "genesis-builder",
     "moved",
     "server",
+    "engine-api",
```
bulenttastan, line 8: nit: alphabetical ordering

RomanHodulak:
Yes, we should move `types/engine_api.rs` to this crate too. The reason I didn't do it here was to keep the diff size manageable for review efficiency.

I'm not in favor of any `moved-types` crate, I don't think that's the way forward. That's too generic and will lead to it being a crate for everything. Just look at aptos-types and how that looks like.

engine-api/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "moved-engine-api"
+description = "Moved execution Ethereum engine JSON-RPC API"
+version = "0.1.0"
```
bulenttastan, line 4: We can rely on the workspace for version and edition, as:
```
version.workspace = true
edition.workspace = true
```

bulenttastan (approved):
Let's merge it after the alphabetical ordering and workspace for version/edition.

birchmd (approved):
Looks good to me, nice work.

Issue #47, opened 2024-08-08
Return op-move response


birchmd:
Correctly implement necessary Ethereum JSON RPC methods

- `eth_estimateGas`
- `eth_getCode`
- `eth_getBalance`
- `eth_transactionReceipt`
- `eth_getBlockByHash`
- `eth_getBlockByHeight`
- etc.

Note: depends on doing #44 and #46 first (and other tasks that remove dependency on op-geth)

Code reference: https://github.com/MovedNetwork/op-move/blob/79df431c0deb8cfe59303d13adff50cbff10ec62/moved/src/main.rs#L211

Generate `TransactionReceipt` using the `op-alloy` create.

bulenttastan:
To prioritize which APIs to start implementing we can look into which ones are called during the integration test.

Without any external interaction these APIs get called by the system:
```
eth_chainId
eth_getBlockByNumber
eth_getBlockByHash
engine_forkchoiceUpdatedV3
engine_getPayloadV3
engine_newPayloadV3
```

Sending a contract deployment and a transaction makes the following additional API calls:
```
eth_getBalance
eth_feeHistory
eth_getTransactionCount
eth_estimateGas
eth_blockNumber
eth_sendRawTransaction
eth_getTransactionReceipt
eth_call
```

PR #137, opened 2024-11-02
feat: Get balance Ethereum API

bulenttastan:
### Description
Returns the requested account balance from `eth_token`. Part of #47

### Changes
- Create a new engine API method
- Open up `eth_token`'s quick balance lookup and use it to implement get balance API

### Testing
- Added a few block number parsing tests since they can be string literal or a hex number ie. `"latest"`, `"0x1"`
- Made sure the API gets called via the integration test and an external client like Postman

### Notes
While testing with the integration test, noticed that the `state_channel` gets closed after a few requests. This is happening at the top level server requests and it happens even without this API. Haven't found a solution yet and looking into this issue. If a solution comes up we can include it in this PR or we create a separate bug issue for it.

### DIFF
--- a/moved/src/move_execution/eth_token.rs
+++ b/moved/src/move_execution/eth_token.rs
@@ -198,7 +198,8 @@ pub fn get_eth_balance<G: GasMeter>(
     }
 }
 
-/// Simplified API for getting the base token balance for use in tests.
+/// Simplified API for getting the base token balance with no side effects.
+/// Use it only for view methods as it does not use a VM session in the request pipeline.
 pub fn quick_get_eth_balance(
     account: &AccountAddress,
     state: &(impl MoveResolver<PartialVMError> + TableResolver),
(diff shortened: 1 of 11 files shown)


moved/src/state_actor.rs
```diff
@@ -171,6 +171,16 @@ impl<
                 self.mem_pool
                     .insert(tx_hash, (ExtendedTxEnvelope::Canonical(tx), encoded));
             }
+            StateMessage::GetBalance {
+                address,
+                response_channel,
+                ..
+            } => {
+                // TODO: Support balance from arbitrary blocks
```
birchmd, line 179: This should be doable via the merkle tree. Look up the block based on the number/tag, get the state root from that block and then use the merkle store to construct a state instance corresponding to that root.

moved/src/move_execution/eth_token.rs
```diff
@@ -198,7 +199,6 @@ pub fn get_eth_balance<G: GasMeter>(
 }
 
 /// Simplified API for getting the base token balance for use in tests.
```
birchmd, line 201: We should update the comment now that this is used outside of tests. Include a warning that this API should not be used in any transaction processing; it should only be used when a VM session does not already exist as part of the request pipeline.

moved/src/state_actor.rs
bulenttastan: Created an issue to tackle this independently: https://github.com/MovedNetwork/op-move/issues/142

PR #139, opened 2024-11-04
feat: Chain Id Ethereum API

bulenttastan:
### Description
Returns the Moved Network chain ID for the `eth_chainId` call. Part of https://github.com/MovedNetwork/op-move/issues/47

### Changes
- Return a static response. There's no need to go through the state channel.

### Testing
- Added a unit test for the `execute` method
- Tested with the integration test. Call to `eth_chainId` returns the chain id in the `op_move_response` as below:
```json
{
  "request": {
    "jsonrpc": "2.0",
    "id": 1,
    "method": "eth_chainId"
  },
  "geth_response": {
    "jsonrpc": "2.0",
    "id": 1,
    "result": "0xa455"
  },
  "op_move_response": {
    "id": 1,
    "jsonrpc": "2.0",
    "result": "0x194"
  },
  "port": "9551"
}
```

### DIFF
--- a/engine-api/src/methods/chain_id.rs
+++ b/engine-api/src/methods/chain_id.rs
@@ -5,7 +5,6 @@ use {
 };
 
 pub async fn execute(
-    _request: serde_json::Value,
     state_channel: mpsc::Sender<StateMessage>,
 ) -> Result<serde_json::Value, JsonRpcError> {
     let response = inner_execute(state_channel).await?;
@@ -52,16 +51,9 @@ mod tests {
         );
         let state_handle = state.spawn();
 
-        let request = serde_json::json!({
-            "id": 30054,
-            "jsonrpc": "2.0",
-            "method": "eth_chainId",
-            "params": []
-        });
-
         let expected_response: serde_json::Value = serde_json::from_str(r#""0x194""#).unwrap();
 
-        let response = execute(request, state_channel).await.unwrap();
+        let response = execute(state_channel).await.unwrap();
 
         assert_eq!(response, expected_response);
         state_handle.await.unwrap();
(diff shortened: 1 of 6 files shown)


engine-api/src/methods/chain_id.rs
```diff
@@ -0,0 +1,41 @@
...
+    crate::jsonrpc::JsonRpcError,
+    moved::{genesis::config::CHAIN_ID, types::state::StateMessage},
+    tokio::sync::mpsc,
+};
+
+pub async fn execute(
+    _request: serde_json::Value,
+    _state_channel: mpsc::Sender<StateMessage>,
+) -> Result<serde_json::Value, JsonRpcError> {
+    Ok(serde_json::to_value(format!("0x{:x}", CHAIN_ID))
```
birchmd, line 11: This should be based on the genesis config in the state actor instead of returning a static response because it will be important to return different chain ids even for our own networks (devnet, testnet, mainnet, must all be different).

It would also be fine to pass the genesis config down to this method instead of asking the state actor for it since the genesis config is static for the whole execution of the chain.

bulenttastan: Getting it from the genesis config via state actor now. Chain id still comes out correct:
```json
"op_move_response": {
    "id": 1,
    "jsonrpc": "2.0",
    "result": "0x194"
}
```

engine-api/src/methods/chain_id.rs
```diff
@@ -0,0 +1,69 @@
+use {
+    crate::{json_utils::access_state_error, jsonrpc::JsonRpcError},
+    moved::types::state::StateMessage,
+    tokio::sync::{mpsc, oneshot},
+};
+
+pub async fn execute(
+    _request: serde_json::Value,
```
RomanHodulak, line 8: No need to pass in the request.

bulenttastan: You're right, removed.

birchmd (approved):

PR #140, opened 2024-11-04
refactor: Split off `engine-api` schema from `moved` to `moved-engine-api` crate

RomanHodulak:
Moves the Engine API request and response bodies into the dedicated crate.

This makes the focus of `moved-engine-api` be the presentation layer and the inner `moved` focused on application and business logic.

Another way to look at this is also that `moved` crate uses "normalized" types that can be created based on multiple versions of the API schema (we currently support only V3 but we will need to support other versions too).


bulenttastan:
Practically the `engine_api.rs` file is split into `schema.rs` and `state.rs` for separation of concerns. Looks good to me.

birchmd (approved):
Looks good to me. I agree separating types used internally from those that are parsed from external sources is a good design pattern.

PR #143, opened 2024-11-06
Feat(vm): EVM native

birchmd:
### Description
Adds a new native to the MoveVM which is an embedded EVM. This will allow us to deploy the L2 contracts that are part of the OP stack without needing to rewrite them. Actually deploying these contracts as part of the genesis process will need to be done in a follow-up PR. Making examples of how to use these contracts for user flows (e.g. withdraw ETH from the L2) will need to be done in another PR after that.

### Changes
- New native included in `create_session`
- New EVM native interface included in our framework bundle

### Testing
- New `test_evm` in `evm_native.rs`

### Notes
There are still quite a few TODOs left as comments. I do not think any of them are blocking for the devnet launch, but we will need to revisit them for testnet and mainnet.


genesis-builder/framework/evm/sources/Evm.move
```diff
@@ -0,0 +1,75 @@
...
+    use aptos_framework::primary_fungible_store::ensure_primary_store_exists;
+    use EthToken::eth_token::get_metadata;
+    use std::error;
+    use std::signer;
+
+    /// For now deploying EVM contracts is restricted to an admin account.
+    /// This restriction may be lifted in the future.
+    const ENOT_OWNER: u64 = 1;
+
+    const OWNER: address = @0x1;
```
bulenttastan, line 12: We can actually define this as a variable in the `Move.toml` and refer to it as `const OWNER: address = @owner;`

moved/src/move_execution/evm_native.rs
```diff
@@ -0,0 +1,862 @@
...
+    smallvec::SmallVec,
+    std::{collections::VecDeque, sync::LazyLock},
+};
+
+const EVM_NATIVE_ADDRESS: AccountAddress = AccountAddress::ONE;
+const EVM_NATIVE_MODULE: &IdentStr = ident_str!("evm");
+const ACCOUNT_STORAGE_LAYOUT: MoveTypeLayout = MoveTypeLayout::U256;
+static CODE_LAYOUT: LazyLock<MoveTypeLayout> =
+    LazyLock::new(|| MoveTypeLayout::Vector(Box::new(MoveTypeLayout::U8)));
+static ACCOUNT_INFO_LAYOUT: LazyLock<MoveTypeLayout> = LazyLock::new(|| {
```
bulenttastan, line 47: I assume this is a struct for `address, value, code`, but it's not obvious. We can add a comment describing this layout.

moved/src/move_execution/evm_native.rs
```diff
@@ -0,0 +1,862 @@
...
+            }
+        };
+        result
+            .add_resource_op(struct_tag, op)
+            .expect("Cannot have duplicate storage index");
+    }
+}
+
+fn account_info_struct_tag(address: &Address) -> StructTag {
+    let name = format!("Info_{}", address.encode_hex());
```
bulenttastan, line 215: We might wanna call this `Account_{}` instead 🤔

genesis-builder/framework/evm/sources/Evm.move
birchmd: Fixed in https://github.com/MovedNetwork/op-move/pull/143/commits/d0d84791b2756844ac7beab0314db5e56d56e451

moved/src/move_execution/evm_native.rs
birchmd: Additional documentation added in https://github.com/MovedNetwork/op-move/pull/143/commits/4e08ce5a1f5b63e9f31ca61f93d8b9012234f49a

moved/src/move_execution/evm_native.rs
birchmd: Changed in https://github.com/MovedNetwork/op-move/pull/143/commits/9cdd48aab42f21335a3b1a48083a716b7efb14fe

moved/src/move_execution/evm_native.rs
```diff
@@ -0,0 +1,862 @@
...
+
+    // Capture changes in native context so that they can be
+    // converted into Move changes when the session is finalized
+    evm_native_ctx.state_changes.push(outcome.state.clone());
+
+    // Commit the changes to the DB so that future Move transactions using
+    // the same session will see them.
+    evm_native_ctx.db.commit(outcome.state);
+
+    let gas_used = EvmGasUsed::new(outcome.result.gas_used());
```
bulenttastan, line 452: We'll probably use how much it costed using Move resources than passing the EVM cost over. But I think this is a good way to start to compare them later to decide what makes more sense.

bulenttastan (approved):
Fantastic work 👏

RomanHodulak (approved):
Great work :feelsgood: 

There are things that could be iterated upon, but I the way I see it this PR is supposed to put it out there and then focus on each part individually as a refactor while there is always a working base.

Some points we could investigate:
* Gas pricing
* Logs as move events to plug into the existing move events to eth logs pipeline
* Storage as resources layout

PR #144, opened 2024-11-06
Fix(transactions): Include script hash in session ID when it is present

birchmd:
### Description
While working on #143 I noticed a TODO left on `SessionId` which should have been fixed as part of #122 . This is an oversight on my part so I am fixing it here.

### Changes
- Include script hash in the `SessionId` when available.

### Testing
- Existing tests

### DIFF
--- a/moved/src/types/session_id.rs
+++ b/moved/src/types/session_id.rs
@@ -32,7 +32,7 @@ impl SessionId {
         tx_hash: &B256,
         genesis_config: &GenesisConfig,
         block_header: HeaderForExecution,
-        script_hash: Option<[u8; 32]>,
+        script_hash: Option<B256>,
     ) -> Self {
         let chain_id = u8_chain_id(genesis_config);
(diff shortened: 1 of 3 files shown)


moved/src/types/session_id.rs
```diff
@@ -29,6 +29,7 @@ impl SessionId {
         maybe_entry_fn: Option<&EntryFunction>,
         tx_hash: &B256,
         genesis_config: &GenesisConfig,
+        script_hash: Option<[u8; 32]>,
```
bulenttastan, line 32: Should we not use `Option<B256>` similar to `tx_hash`?

birchmd: It means having an extra `map` involved because the aptos library where this ultimately gets used does not know about the alloy types, but I agree being consistent makes sense. Fixed in https://github.com/MovedNetwork/op-move/pull/144/commits/df13715d3eb74708cbd946ab567b3a3b19ff4cd8

PR #146, opened 2024-11-11
feat(engine-api): Define JSON-RPC API schema and methods for block fetching

RomanHodulak:
#47


engine-api/src/schema/block.rs
```diff
@@ -0,0 +1,93 @@
...
+    moved::{
+        primitives::{Address, Bytes, B2048, B256},
+        types::state::{BlockResponse, TransactionResponse},
+    },
+    serde::{Deserialize, Serialize},
+};
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GetBlockResponse {
```
bulenttastan, line 11: According to the docs `baseFeePerGas` and `mixHash` parameters are missing.
https://www.quicknode.com/docs/optimism/eth_getBlockByHash

When I made a call to this API while running the integration test, I see the following params are missing:
`baseFeePerGas`, `blobGasUsed`, `excessBlobGas`, `mixHash`, `parentBeaconBlockRoot`, `withdrawals`, `withdrawalsRoot`

engine-api/src/schema/block.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// the block number. null when its pending block.
```
bulenttastan, line 12: nit: We usually capitalize the comments, ie `/// The ...`
Applies to all of them below.

moved/src/types/state.rs
```diff
@@ -102,6 +103,85 @@ pub enum StateMessage {
     },
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BlockResponse {
+    /// the block number. null when its pending block.
```
bulenttastan, line 108: nit: same comment capitalization `/// The ...`

engine-api/src/request.rs
```diff
@@ -52,5 +52,7 @@ async fn inner_handle_request(
         ForkChoiceUpdatedV2 => todo!(),
         GetPayloadV2 => todo!(),
         NewPayloadV2 => todo!(),
+        GetBlockByHash => get_block_by_hash::execute(request, state_channel).await,
```
bulenttastan, line 55: nit: We can keep these above the `todo!()` APIs

engine-api/src/schema/block.rs
birchmd: Is there a type from the alloy library that can be used here instead of defining it ourselves?

moved/src/types/state.rs
birchmd: Is there a type from the alloy library that can be used here instead of defining it ourselves?

RomanHodulak: `alloy::consensus::Block`

engine-api/src/schema/block.rs
bulenttastan: I think [alloy-rpc-types-eth](https://github.com/alloy-rs/alloy/tree/main/crates/rpc-types-eth) crate has a lot of the receipt structs.

RomanHodulak: I think that receipt struct is specific to native Ethereum but needs to be extended with Optimism specific fields.

engine-api/src/schema/eth/block.rs
```diff
@@ -0,0 +1,16 @@
+pub use alloy::eips::BlockNumberOrTag;
+
+use {
+    alloy::rpc,
+    moved::types::state::BlockResponse,
+    serde::{Deserialize, Serialize},
+};
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct GetBlockResponse(rpc::types::Block<rpc::types::Transaction>);
```
birchmd, line 10:
```suggestion
pub struct GetBlockResponse(pub rpc::types::Block<rpc::types::Transaction>);
```

Maybe want to make the inner value `pub` in case we want to use it somewhere in the code.

moved/src/types/state.rs
```diff
@@ -102,6 +105,52 @@ pub enum StateMessage {
...
+            // TODO: review fields below
+            uncles: Vec::new(),
+            size: None,
+            withdrawals: None,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TransactionResponse {
```
birchmd, line 149: What is this going to be used for?

birchmd (approved):
Good work!

engine-api/src/schema/eth/mod.rs
```diff
@@ -0,0 +1,4 @@
+mod block;
+mod withdrawal;
+
+pub use {block::*, withdrawal::*};
```
bulenttastan, line 4: nit: We usually keep `pub use` at the top.

engine-api/src/schema/mod.rs
```diff
@@ -0,0 +1,4 @@
+mod engine;
+mod eth;
+
+pub use {engine::*, eth::*};
```
bulenttastan, line 4: nit: same here

PR #147, opened 2024-11-12
Fix(integration-test): Attempt to clean all files instead of bailing

birchmd:
### Description
Running `cleanup_files` at the start of the integration test is meant to make sure it is starting from a good state, even if it failed at a random point previously. This means that some of the files may not exist and therefore we want to simply ignore all errors related to deleting those files instead of using `?` to exit the function early.

### Testing
- Integration test


bulenttastan (approved):
This is what I was looking for. Thanks for fixing it 🙌

PR #148, opened 2024-11-12
Feat(genesis): Deploy L2 EVM contracts at genesis

birchmd:
### Description
This PR extends the genesis process to include deploying the EVM contracts coming from OP stack. These contracts are taken directly from the `genesis.json` file created by `op-node`. For the purpose of testing in the `moved` module, a new file is added to `moved/src/tests/res`. This file comes from the genesis file generated by the integration test. The integration test itself uses the file it generates.

The reason to re-use the same file to initialize op-move as is used for op-geth is because we want op-move to essentially be a drop-in replacement for op-geth.

### Changes
- OP stack L2 EVM contracts included in Moved genesis state

### Testing
Existing tests (including integration test)

### Notes
There are a few new TODOs added in this PR, but they are P2 (not devnet blocking) in my opinion.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6457,6 +6457,7 @@ dependencies = [
  "once_cell",
  "revm",
  "serde",
+ "serde_json",
  "sha2 0.10.8",
  "smallvec",
  "sui-framework",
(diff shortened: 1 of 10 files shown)


RomanHodulak (approved):
Nice work

PR #149, opened 2024-11-13
Chore(evm-native): Split into multiple modules

birchmd:
### Description
This PR does not change any implementation, it simply splits the large `evm_native.rs` into smaller modules. This change is motivated by an idea I have to include an additional native function for solidity ABI encoding (it should help with the Move<>EVM interface inside move contracts). Adding mote logic to the one evm_native file didn't seem easy to maintain, so I split it out into multiple files. Now when I make the new native in a future PR, it can have most of its logic in a file of its own.

### Changes
N/A

### Testing
Existing tests.


moved/src/move_execution/evm_native/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+/// across all the contracts using that bytecode.
+static ACCOUNT_INFO_LAYOUT: LazyLock<MoveTypeLayout> = LazyLock::new(|| {
+    MoveTypeLayout::Struct(MoveStructLayout::Runtime(vec![
+        MoveTypeLayout::U256,                                 // balance
+        MoveTypeLayout::U64,                                  // nonce
+        MoveTypeLayout::Vector(Box::new(MoveTypeLayout::U8)), // code_hash
+    ]))
+});
+
+pub use {
```
bulenttastan, line 46: Should this be at the top? Is it more common to keep `pub use` statements at the bottom?
In other files I see it being the first line, even above the non-pub `use`.

bulenttastan:
The folder seems quite isolated from `move_execution` to even become a separate crate (except `eth_native/tests.rs`).
What do you think about moving it to a crate?

moved/src/move_execution/evm_native/mod.rs
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/149/commits/e55b7412c36c436fb77c7e313cf477ba76ccb904

birchmd:
We could move the native to a separate crate and leave the test in `moved` crate. But unless we plan on re-using the native elsewhere or publishing it to crates.io separately I don't see a strong reason to do so. I like that the test can be encapsulated with the rest of the evm_native code when it is part of the main crate.

bulenttastan (approved):

PR #150, opened 2024-11-13
Feat(evm-native): Implement native function for Solidity ABI encoding

birchmd:
### Description
This PR adds a new native function for encoding arbitrary move types to Solidity ABI format. The motivation for this is to make it easier to implement Move interface for EVM-based contracts. You can see the example of an ERC-20 transfer in the new test that is part of this PR.

### Changes
- New native function for Solidity ABI encoding in evm Move module

### Testing
- Extended `test_evm` to include an ERC-20 transfer triggered from a Move interface where the EVM arguments are encoded using the new native.

### Notes
I am not charging gas for this new native. I don't think this is a big deal in the short term, but definitely something we need to look into as part of the other issue about gas usage in the EVM native.


genesis-builder/framework/evm/sources/Evm.move
```diff
@@ -58,6 +70,12 @@ module Evm::evm {
         native_evm_create(caller_addr, get_asset_value(value), data)
     }
 
+    /// Encode the move value into bytes using the Solidity ABI
+    /// such that it would be suitable for passing to a Solidity contract's function.
+    /// The prefix can be used to prepend the output with a Solidity 4-byte function
+    /// selector if needed.
+    public native fun abi_encode_params<T>(prefix: vector<u8>, value: T): vector<u8>;
```
bulenttastan, line 77: Really nice!
Just curious about 2 things (1) Do we create a separate native impl, even though it's part of the same Evm module, so that we can isolate Move to EVM calls? (2) Is this considered the infrastructure for cross-contract calls?

birchmd: (1) It could be a separate native impl since it does not use any of the EVM machinery; it is just a pure function. But on the other hand, I'm not sure why a MoveVM instance would want this native without having the other EVM natives also. I cannot think of a reason why a Move contract would want to use this encoding without then also calling the EVM.

Having be another function in the same native is a little less boilerplate, so unless we have a particular reason to separate the encoding out, I think it is fine as-is.

(2) This is infrastructure to make it easier to call EVM contracts from Move contracts. We'll use it ourselves to call the OP stack L2 contracts. And if we open up this EVM to all users in the future then it will also be useful in general.

moved/src/move_execution/evm_native/solidity_abi.rs
```diff
@@ -0,0 +1,178 @@
...
+            )
+        }
+    }
+}
+
+fn force_to_u8(mv: MoveValue) -> u8 {
+    if let MoveValue::U8(x) = mv {
+        x
+    } else {
+        0
```
bulenttastan, line 176: Should maybe `panic!` here? 🤔

bulenttastan (approved):

PR #151, opened 2024-11-18
U256 ETH fungible asset balances

### LINKED ISSUE #56: U256 token balances
The Ethereum standard for token balances (ETH, ERC20) is to use the U256 data type. In the Aptos Move framework though it is u64. We probably need to write our own token standard in Move that uses U256 balances.

Code reference:
https://github.com/MovedNetwork/op-move/blob/79df431c0deb8cfe59303d13adff50cbff10ec62/moved/src/move_execution.rs#L69

bulenttastan:
### Description
Support for `u256` token balance for native ETH token that will be used throughout the chain.
Created a `u256` amount version of `fungible_asset` and `primary_fungible_store` modules under `aptos-framework`.
Closes #56 

### Changes
- Duplicate the 2 modules by changing the amounts to `u256`, and removing the support for concurrency and dynamic function calls from the original implementations.
- Update the genesis file generation to include these modules and add `primary_fungible_store_u256` as a friend to the `object.move` module.
- Update tests to use `u256` amounts instead of `u64`.

### Testing
✓ Added a Move test in eth-token' module
✓ Updated the unit tests, especially move execution tests
✓ Integration test still passes

### Notes
The only injection to the existing framework is by adding the `primary_fungible_store_u256` as a friend to the `object.move` module. Object module already friends the original `primary_fungible_store` module and the injection is automated in the genesis builder, so it's not done manually.


genesis-builder/src/main.rs
```diff
@@ -126,6 +126,22 @@ fn fix_aptos_packages() -> anyhow::Result<()> {
...
+
+    // Object module should also friend the new u256 version of the fungible store
+    let object_move_file = &APTOS_FRAMEWORK_DIR
+        .join("aptos-framework")
+        .join("sources")
+        .join("object.move");
+    let mut content = read_to_string(object_move_file)?;
+    content = content.replace(
+        "friend aptos_framework::primary_fungible_store;",
+        "friend aptos_framework::primary_fungible_store;\n    friend aptos_framework::primary_fungible_store_u256;"
```
bulenttastan, line 138: Any better way to write this line?

engine-api/src/methods/get_payload.rs
```diff
@@ -159,7 +159,7 @@ mod tests {
                 "executionPayload": {
                     "parentHash": "0xe56ec7ba741931e8c55b7f654a6e56ed61cf8b8279bf5e3ef6ac86a11eb33a9d",
                     "feeRecipient": "0x4200000000000000000000000000000000000011",
-                    "stateRoot": "0xf3a022e9d83ae2c38348e3b726a1ec25a9d6a5c2a15913d187e8beaa1c3b5d7d",
+                    "stateRoot": "0x0e630edda6cec7ad91c29ba4573bca730cebcea96a7d100bc4d800f12f5a8140",
```
bulenttastan, line 162: Is this the right thing to do, to change the state root to what it returns when you run the test?

moved/src/move_execution/tests.rs
```diff
@@ -1523,6 +1526,11 @@ impl ScriptCompileJob {
             let mut framework = aptos_framework::testnet_release_bundle()
                 .files()
                 .expect("Must be able to find Aptos Framework files");
+            let genesis_base = "../genesis-builder/framework/aptos-framework/sources";
```
bulenttastan, line 1529: Renaming `genesis_base` to `aptos_framework_base`

genesis-builder/src/main.rs
bulenttastan: Alternatively the following, which is at least not as long:
```rust
content = content.replace(
        "friend aptos_framework::primary_fungible_store;",
        "friend aptos_framework::primary_fungible_store;
    friend aptos_framework::primary_fungible_store_u256;"
```

engine-api/src/methods/get_payload.rs
birchmd: Yes, the state root needs to be updated whenever we change the framework. This is the right thing to do.

genesis-builder/src/main.rs
birchmd: I think this is ok as is. The version with the line break would also be fine if you prefer it.

birchmd (approved):
Nice job!

PR #153, opened 2024-11-19
Feat(transactions): Withdrawal flow

birchmd:
### Description
This is the first step towards supporting the [OP stack withdrawal flow](https://docs.optimism.io/stack/transactions/withdrawal-flow).

### Changes
- New Move module in our framework representing the L2CrossDomainMessenger contract from OP stack
- Updates to the Evm Move module to allow checking for EVM execution success and including the EVM logs in the Move logs
- Fix bug in genesis where the storage for the L2 contracts in the EVM was not being initialized properly
- Fix bug where EVM native was not working properly with non-zero attached ETH
- Refactor Log and ToLog traits to handle EVM logs properly

### Testing
- New test `test_withdrawal_tx`

### Notes
There is no integration test yet to actually see the tokens show up on the L1, but I should make one after we are using op-move responses instead of geth responses in the integration test. To make that work there will likely need to be changes to the L1 contract since our Merkle proof will be a little different, but we can cross that bridge when we come to it.


moved/src/genesis/framework.rs
```diff
@@ -49,6 +52,16 @@ static SUI_SYSTEM_PACKAGES: Lazy<BTreeMap<ObjectID, SystemPackage>> = Lazy::new(
     snapshot.into_iter().map(|pkg| (pkg.id, pkg)).collect()
 });
 
+const fn eth_address(value: &[u8; 20]) -> AccountAddress {
+    let mut buf = [0u8; 32];
+    let mut i = 0;
+    while i < 20 {
```
bulenttastan, line 58: nit: `for i in 0..20 {` might be easier to read

birchmd: `for` loops are currently not allowed in `const` functions unfortunately.

moved/src/move_execution/tests.rs
```diff
@@ -560,6 +590,74 @@ fn test_deposit_tx() {
...
 }
 
+#[test]
+fn test_withdrawal_tx() {
+    let genesis_config = GenesisConfig::default();
+    let mut signer = Signer::new(&PRIVATE_KEY);
+    let (_, mut state) = deploy_contract("natives", &mut signer, &genesis_config);
+
+    // 1. Deposit ETH to user
+    let mint_amount = U256::from(123u64);
```
bulenttastan, line 600: We can make filling up the user's account a function, just to have some balance It's used in a few places now.

moved/src/move_execution/tests.rs
```diff
@@ -1475,6 +1573,10 @@ impl ModuleCompileJob {
                 "evm_admin".into(),
                 NumericalAddress::parse_str("0x1").unwrap(),
             ),
+            (
```
bulenttastan, line 1576: We can create a convenience function just like `add_custom_framework_paths` to include these named addresses.

bulenttastan (approved):
Looks really good.
Left some comments for convenience methods.

moved/src/move_execution/tests.rs
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/153/commits/d1d5cb6e09fa9a5f5fba22fe36170d0953b93d62

moved/src/move_execution/tests.rs
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/153/commits/eb4db738d998b6b6de0dddbc7359b5d9fc0aa703

bulenttastan (approved):

genesis-builder/framework/evm/sources/Evm.move
```diff
@@ -15,13 +16,17 @@ module Evm::evm {
...
-    /// TODO: what capabilities should this have?
-    struct EvmLog has drop {
+    struct EvmLog has copy, store, drop {
         addr: address,
         topics: vector<u256>,
         data: vector<u8>,
     }
 
+    #[event]
+    struct EvmLogsEvent has store, drop {
```
RomanHodulak, line 26: Why is it wrapped like that?

RomanHodulak (approved):
Excellent progress :chart_with_upwards_trend:

genesis-builder/framework/evm/sources/Evm.move
birchmd: Honestly I am not sure what it does, but I [saw it in the Aptos framework](https://github.com/aptos-labs/aptos-core/blob/aptos-node-v1.14.0/aptos-move/framework/aptos-framework/sources/fungible_asset.move#L185) as a way of marking structs you wanted to emit as events, so I did it too.

PR #155, opened 2024-11-20
refactor: Test utility and modular test files

bulenttastan:
### Description
This merely refactors existing `move_execution` tests by moving tests into separate files per Move module and creates a test utility library to handle common Move module operations.

### Changes
- Utility library for common test operations
- Tests are moved to individual modules

### Testing
✓ Unit tests and integration test pass


moved/src/move_execution/tests/utils.rs
```diff
@@ -0,0 +1,631 @@
...
+    }
+
+    /// Transfers ETH to a specified address with L1 cost considerations
+    ///
+    /// # Arguments
+    /// * `to` - Destination address for the transfer
+    /// * `amount` - Amount of ETH to transfer
+    /// * `l1_cost` - L1 cost associated with the transaction
+    /// * `expect_error` - Whether the transfer is expected to fail
+    pub fn transfer(&mut self, to: Address, amount: u64, l1_cost: u64, expect_error: bool) {
```
birchmd, line 134: Nit: passing a Boolean argument for internal control flow doesn't end up looking very intuitive at the call site. A better way to approach this is probably to return `outcome.vm_outcome` from this function so that the caller can decide to call `unwrap` or `unwap_err` as needed.

moved/src/move_execution/tests/utils.rs
```diff
@@ -0,0 +1,631 @@
...
+    /// Executes a Move entry function with the given arguments
+    ///
+    /// This is the recommended way to call Move functions in tests as it handles
+    /// argument serialization and transaction creation.
+    ///
+    /// # Arguments
+    /// * `module_id` - The ModuleId containing the function to execute
+    /// * `function` - Name of the function to call
+    /// * `args` - Vector of Move values to pass as arguments
+    pub fn execute(&mut self, module_id: &ModuleId, function: &str, args: Vec<&MoveValue>) {
```
birchmd, line 168: Nit: there's no reason for `args` to be `Vec`. And type implementing `IntoIterator<Item=&MoveValue>` would work (for example an array instead of `Vec`). Having the implementation be generic could avoid some unnecessary allocations.

birchmd (approved):
Looks good! I have a couple of nits, but nothing blocking.

PR #156, opened 2024-11-21
feat: Implement block fetching queries and integrate with JSON-RPC API methods

RomanHodulak:
- Separate commands and queries in `StateActor`
- Introduce new `BlockQueries` trait specialized for dealing with block fetching
- Share context with both `BlockRepository` and `BlockQuery` to avoid introducing locks by moving the ownership one layer above

Part of #47

### DIFF
--- a/moved/src/block/root.rs
+++ b/moved/src/block/root.rs
@@ -1,15 +1,20 @@
 use {
     crate::{
         primitives::{Address, Bytes, B2048, B256, U256},
-        types::transactions::ExtendedTxEnvelope,
+        types::{state::BlockResponse, transactions::ExtendedTxEnvelope},
     },
     alloy::{primitives::hex, rlp::RlpEncodable},
     std::fmt::Debug,
 };
 
-pub trait BlockRepository: Debug {
-    fn add(&mut self, block: ExtendedBlock);
-    fn by_hash(&self, hash: B256) -> Option<ExtendedBlock>;
+pub trait BlockQueries<Storage>: Debug {
+    fn by_hash(&self, storage: &Storage, hash: B256) -> Option<BlockResponse>;
+    fn by_height(&self, storage: &Storage, height: u64) -> Option<BlockResponse>;
+}
+
+pub trait BlockRepository<Storage>: Debug {
+    fn add(&mut self, storage: &mut Storage, block: ExtendedBlock);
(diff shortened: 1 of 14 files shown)


moved/src/block/root.rs
```diff
@@ -1,15 +1,20 @@
...
     },
     alloy::{primitives::hex, rlp::RlpEncodable},
     std::fmt::Debug,
 };
 
-pub trait BlockRepository: Debug {
-    fn add(&mut self, block: ExtendedBlock);
-    fn by_hash(&self, hash: B256) -> Option<ExtendedBlock>;
+pub trait BlockQueries<Context>: Debug {
+    fn by_hash(&self, context: &Context, hash: B256) -> Option<BlockResponse>;
```
birchmd, line 11: Nit: I don't like the name `context` for this variable and type generic. In the implementation it's the backing store, which is just in memory for now, but presumably we'll have a persistent store in the future. I think a better name would reflect that this type represents the physical storage.

birchmd (approved):
Nice work! I like working in the query/command separation here. I think it lays some nice ground work for larger refactoring in the future.

moved/src/state_actor.rs
```diff
@@ -117,27 +132,64 @@ impl<
...
+        match msg {
+            Query::ChainId { response_channel } => {
+                response_channel.send(self.genesis_config.chain_id).ok()
+            }
+            Query::GetBalance {
+                address,
+                response_channel,
+                ..
+            } => {
+                // TODO: Support balance from arbitrary blocks
```
bulenttastan, line 150: For a separate PR, but we can get this `TODO` item implemented now with the `block_memory`, right?

bulenttastan (approved):

birchmd (approved):

moved/src/block/root.rs
RomanHodulak: I have a few suggestions such as:
* Storage
* PhysicalStorage
* BackingStore
* One of the above with suffix Access

I went with `Storage` for the generic and `storage` for the variable.

You also said you don't like the type generic. Did you mean you don't like the name of the generic parameter `Context` or something else?

bulenttastan: I'll add a few suggestions too:
- BlockStore
- BlockStorage

birchmd: I just meant the name `Context`. I like `Storage` fine.

PR #157, opened 2024-11-21
Chore(deps): Upgrade alloy to version 0.6

birchmd:
### Description
Keeping the `alloy` dependency up to date. This will be useful for when we start using `op-alloy`.

### Changes
- alloy 0.5 -> alloy 0.6

### Testing
Existing tests


moved/src/types/state.rs
```diff
@@ -111,34 +111,37 @@ impl From<ExtendedBlock> for BlockResponse {
...
+                    transactions_root: value.block.header.transactions_root,
+                    state_root: value.block.header.state_root,
+                    receipts_root: value.block.header.receipts_root,
+                    difficulty: value.block.header.difficulty,
+                    extra_data: value.block.header.extra_data,
+                    gas_limit: value.block.header.gas_limit,
+                    gas_used: value.block.header.gas_used,
+                    timestamp: value.block.header.timestamp,
+                    beneficiary: value.block.header.beneficiary,
+                    // TODO: review fields below
```
bulenttastan, line 133: Just curious, can we review this `TODO` by calling the integration test?

bulenttastan (approved):

RomanHodulak (approved):

moved/src/types/state.rs
birchmd: Yeah, the integration test might help by letting us look at real, fully populated values coming from `op-geth`. We should also review the documentation to check our understanding of what the fields mean.

PR #158, opened 2024-11-21
Feat(rpc): implementation for eth_getTransactionCount

birchmd:
### Description
Implementation for `eth_getTransactionCount` RPC method in op-move.

### Changes
- New Query type for getting account nonces
- New quick_get_nonce API

### Testing
- New test in the same style as other RPC modules


RomanHodulak (approved):
Nice one :feelsgood:

Issue #160, opened 2024-11-25
Charge gas for computation on L2


birchmd:
- Compute effective gas price
- Charge user effective gas price * gas limit at tx start
- Refund effective gas price * (gas limit - gas used) at tx end

lkilu6456546:
Hello, I tried to solve the issue.

This is what I did:

Added effective gas price calculation, initial gas charge, and unused gas refund in the execute_canonical_transaction function. Modified the TransactionExecutionOutcome struct to include refund information.

You can review changes in this commit: https://github.com/lkilu6456546/MovedNetwork-op-move-160/commit/fe1493498297d7dffed9c5928d77b9cd50a368db.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

PR #162, opened 2024-11-25
feat(eth-api): Implement `include_transactions` parameter for block fetching

RomanHodulak:
### Description
Implement `include_transactions` parameter for block fetching

### Changes
- Converts the transactions into its full RPC representation
- Computes hashes if full transactions are not included

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### DIFF
--- a/moved/src/types/state.rs
+++ b/moved/src/types/state.rs
@@ -168,45 +168,98 @@ impl From<Query> for StateMessage {
     }
 }
 
-pub type BlockResponse = alloy::rpc::types::Block<alloy::rpc::types::Transaction>;
+pub type RpcBlock = alloy::rpc::types::Block<alloy::rpc::types::Transaction<ExtendedTxEnvelope>>;
 
-impl From<ExtendedBlock> for BlockResponse {
-    fn from(value: ExtendedBlock) -> Self {
-        Self {
+#[derive(Debug)]
+pub struct BlockResponse(pub RpcBlock);
+
+impl BlockResponse {
+    fn new(value: RpcBlock) -> Self {
+        Self(value)
+    }
+
+    pub fn from_block_with_transaction_hashes(value: ExtendedBlock) -> Self {
+        Self::new(RpcBlock {
+            transactions: BlockTransactions::Hashes(
+                value
+                    .block
+                    .transactions
+                    .iter()
+                    .map(ExtendedTxEnvelope::compute_hash)
+                    .collect(),
+            ),
             header: alloy::rpc::types::Header {
                 hash: value.hash,
-                inner: alloy::consensus::Header {
-                    number: value.block.header.number,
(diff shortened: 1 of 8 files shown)


moved/src/types/state.rs
```diff
@@ -159,45 +160,97 @@ impl From<Query> for StateMessage {
...
+                    .block
+                    .transactions
+                    .into_iter()
+                    .filter_map(ExtendedTxEnvelope::into_canonical)
+                    .enumerate() // TODO: should index count deposited tx?
+                    .map(|(i, inner)| alloy::rpc::types::Transaction {
+                        block_hash: Some(value.hash),
+                        block_number: Some(value.block.header.number),
+                        transaction_index: Some(i as u64),
+                        effective_gas_price: None,
```
birchmd, line 207: Worth adding a TODO here referencing #160

RomanHodulak: I'm not sure, the doc comment said it's deprecated.

moved/src/types/transactions.rs
```diff
@@ -29,12 +34,149 @@ pub struct DepositedTx {
...
+    pub fn sender(&self) -> Option<Address> {
+        match self {
+            Self::Canonical(tx) => tx.recover_signer().ok(),
+            Self::DepositedTx(tx) => Some(tx.from),
+        }
+    }
+}
+
+impl TransactionTrait for ExtendedTxEnvelope {
+    fn chain_id(&self) -> Option<ChainId> {
```
bulenttastan, line 60: I hate macros, but these seem so macro-able as single line macros. Just a thought.

moved/src/types/state.rs
birchmd: Yes, I see. The [PR where the field was introduced](https://github.com/alloy-rs/alloy/pull/1630) says that it is only used when the inner transaction type does not have its own gas price set. I that means we will still need to set this field in some circumstances because `TxEip1559` has `fn gas_price() -> None` in its implementation of `Transaction`.

moved/src/types/transactions.rs
```diff
@@ -29,12 +34,129 @@ pub struct DepositedTx {
...
+    fn nonce(&self) -> u64 {
+        0
+    }
+
+    fn gas_limit(&self) -> u64 {
+        self.gas.to_u64()
+    }
+
+    fn gas_price(&self) -> Option<u128> {
+        Some(0)
```
birchmd, line 104: I think should be `None` instead

moved/src/types/state.rs
RomanHodulak: Good find.

birchmd (approved):
Nice work!

PR #163, opened 2024-11-25
Feat(rpc): eth_getTransactionReceipt implementation

birchmd:
### Description
Implementation for the `eth_getTransactionReceipt` RPC method. Several changes were needed to have all the data needed to populate the various fields.

### Changes
- Added `op-alloy` dependency to get access to OP stack types.
- New field `tx_receipts` in state actor to keep track of transaction receipts (and what block the transaction was executed in).
- Keep track of transaction index and log index in `execute_transactions`
- Record new module deployment in `TransactionExecutionOutcome`
- New query type for `eth_getTransactionReceipt`
- New module `get_transaction_receipt` to send query to state actor

### Testing
- New tests in `engine-api/src/methods/get_transaction_receipt.rs`

### Notes
- The `effective_gas_price` is left as 0 in this PR because we currently don't charge gas. This should be fixed as part of #160 
- There is still a TODO about `deposit_nonce` and `deposit_receipt_version` fields. We'll either need to look at op-geth or the OP stack documentation to figure out what these should be. This can be done in a future PR.


RomanHodulak (approved):

engine-api/src/methods/forkchoice_updated.rs
```diff
@@ -134,6 +138,39 @@ pub(super) mod tests {
         "#).unwrap()
     }
 
+    pub fn create_state_actor() -> (moved::state_actor::InMemStateActor, Sender<StateMessage>) {
```
bulenttastan, line 141: We could move this to higher level `mod.rs` to use it for all API tests.

PR #164, opened 2024-11-28
feat(eth-api): Support block fetching by tag

RomanHodulak:
### Description
Implement `tag` parameter for block fetching

### Changes
- Adds support of block fetching by tag

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt


moved/src/state_actor.rs
```diff
@@ -193,9 +193,10 @@ impl<
                 response_channel,
                 include_transactions,
             } => response_channel.send(match height {
-                BlockNumberOrTag::Number(height) => self.block_queries.by_height(&self.block_memory, height, include_transactions),
-                // TODO: Support block "tag"
-                _ => None,
+                Number(height) => self.block_queries.by_height(&self.block_memory, height, include_transactions),
+                Latest | Finalized | Safe => self.block_queries.by_height(&self.block_memory, self.height, include_transactions),
+                Earliest => self.block_queries.by_height(&self.block_memory, 0, include_transactions),
+                Pending => None,
```
bulenttastan, line 199: I don't know we should support this, but my understanding was `Pending` meant the transactions in the `mem_pool`.

moved/src/state_actor.rs
```diff
@@ -193,9 +193,10 @@ impl<
                 response_channel,
                 include_transactions,
             } => response_channel.send(match height {
-                BlockNumberOrTag::Number(height) => self.block_queries.by_height(&self.block_memory, height, include_transactions),
-                // TODO: Support block "tag"
-                _ => None,
+                Number(height) => self.block_queries.by_height(&self.block_memory, height, include_transactions),
```
bulenttastan, line 196: Is it easy to write positive and negative tests for this?

moved/src/state_actor.rs
```diff
@@ -507,3 +507,6 @@ impl<
         Some(result)
     }
 }
+
+#[cfg(test)]
+mod tests;
```
bulenttastan, line 512: We should keep this at the beginning

moved/src/state_actor/tests.rs
```diff
@@ -0,0 +1,130 @@
+use {
```
bulenttastan, line 1: These tests are really good, but I think we should move them inside `get_block_by_number` method.

bulenttastan: Pushed this change

bulenttastan (approved):

PR #166, opened 2024-12-02
Feat(rpc): Simulate transaction for gas estimation

bulenttastan:
### Description
Accept real estimate gas requests by simulating the transaction and returning the gas used.

### Changes
- Simulate a transaction in `state_actor`.
- Use real module code as input, taken from SDK counter example.
- Normalize transaction request from user.
- Update `alloy` features to support random B256 generation.

### Testing
✓ Unit test is updated with a real deployment example.

### DIFF
--- a/engine-api/src/json_utils.rs
+++ b/engine-api/src/json_utils.rs
@@ -30,3 +30,7 @@ pub fn deserialize<T: DeserializeOwned>(x: &serde_json::Value) -> Result<T, Json
 pub fn access_state_error<E: fmt::Debug>(e: E) -> JsonRpcError {
     JsonRpcError::access_state_error(e)
 }
+
+pub fn transaction_error<E: fmt::Debug>(e: E) -> JsonRpcError {
+    JsonRpcError::without_data(32603, format!("Invalid transaction: {e:?}"))
+}
--- a/engine-api/src/json_utils.rs
+++ b/engine-api/src/json_utils.rs
@@ -32,5 +32,5 @@ pub fn access_state_error<E: fmt::Debug>(e: E) -> JsonRpcError {
 }
 
 pub fn transaction_error<E: fmt::Debug>(e: E) -> JsonRpcError {
-    JsonRpcError::without_data(32603, format!("Invalid transaction: {e:?}"))
+    JsonRpcError::without_data(3, format!("Execution reverted: {e:?}"))
 }
--- a/moved/src/state_actor.rs
+++ b/moved/src/state_actor.rs
@@ -20,7 +20,10 @@ use {
                 StateMessage, ToPayloadIdInput, TransactionReceipt, TransactionWithReceipt,
                 WithExecutionOutcome, WithPayloadAttributes,
             },
-            transactions::{ExtendedTxEnvelope, NormalizedExtendedTxEnvelope},
+            transactions::{
+                ExtendedTxEnvelope, NormalizedEthTransaction, NormalizedExtendedTxEnvelope,
+                TransactionExecutionOutcome,
+            },
         },
         Error::{InvalidTransaction, InvariantViolation, User},
     },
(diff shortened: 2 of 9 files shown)


moved/src/state_actor.rs
```diff
@@ -450,6 +460,36 @@ impl<
         (outcome, receipts)
     }
 
+    fn simulate_transaction(&self, request: TransactionRequest) -> TransactionExecutionOutcome {
+        let mut tx = NormalizedEthTransaction::from(request.clone());
+        if let Some(from) = request.from {
+            tx.nonce = quick_get_nonce(&from.to_move_address(), self.state.resolver());
```
birchmd, line 466: I think we should only set the nonce if `request.from.is_some()` and `request.nonce.is_none()`. If the user gave us a nonce then we should use it.

moved/src/state_actor.rs
```diff
@@ -450,6 +460,36 @@ impl<
...
+            &tx,
+            &B256::random(),
+            self.state.resolver(),
+            &self.genesis_config,
+            0,
+            &(),
+            block_header,
+        ) {
+            Ok(outcome) => outcome,
+            Err(_) => unreachable!("User errors are handled in execution"),
```
birchmd, line 489: Errors other than user errors are possible here. For example, if the user provides their own gas limit then they could theoretically trigger `InvalidTransaction(InvalidTransactionCause::InsufficientIntrinsicGas)`.

I think this function should return a result and the `Err` case is handled in the JSON RPC by returning an error in the response.

moved/src/state_actor.rs
```diff
@@ -450,6 +460,36 @@ impl<
...
+            prev_randao: B256::random(),
+        };
+
+        match execute_transaction(
+            &tx,
+            &B256::random(),
+            self.state.resolver(),
+            &self.genesis_config,
+            0,
+            &(),
```
birchmd, line 485: We need the real impl for `BaseTokenAccounts` here otherwise the simulation will not correctly transfer tokens. Use `&self.base_token`.

bulenttastan: The issue with using `&self.base_token` is that if the user doesn't have any tokens it fails on `transfer_eth` even just to get a transaction estimate. Not sure if that's an expected behavior, so left it with `&()` to ignore any token transfer. What's the good compromise here?

birchmd: I think getting an error from `eth_estimateGas` when you have insufficient balance is expected behaviour. You could give it a try on an Ethereum testnet if you want.

moved/src/state_actor.rs
bulenttastan: Returning the error now, but as it is.
Should we respond with a standard message and ignore the actual internal error message?

engine-api/src/json_utils.rs
```diff
@@ -30,3 +30,7 @@ pub fn deserialize<T: DeserializeOwned>(x: &serde_json::Value) -> Result<T, Json
 pub fn access_state_error<E: fmt::Debug>(e: E) -> JsonRpcError {
     JsonRpcError::access_state_error(e)
 }
+
+pub fn transaction_error<E: fmt::Debug>(e: E) -> JsonRpcError {
+    JsonRpcError::without_data(32603, format!("Invalid transaction: {e:?}"))
```
birchmd, line 35: Is this status code what is returned by geth for this class of errors?

birchmd (approved):

engine-api/src/json_utils.rs
bulenttastan: I got the error code from Ethereum docs: https://ethereum-json-rpc.com/errors

When I send an invalid transaction to `geth`, I get the following error message:
```json
"error": {
    "code": 3,
    "message": "execution reverted",
    "data": "0x000000000000005e"
}
```

So seems like we can return error code `3` with a static `Execution reverted` message.

PR #167, opened 2024-12-02
refactor: Merge the calls to block fetching by number as it only differs in the block number

RomanHodulak:
### Description
In block fetching, the query and the function call and all other parameters except height are always the same. For better readibility, this change rearranges the code to reflect that fact.

### Changes
- Merge the calls to block fetching by number as it only differs in the block number

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### DIFF
--- a/moved/src/state_actor.rs
+++ b/moved/src/state_actor.rs
@@ -192,11 +192,14 @@ impl<
                 height,
                 response_channel,
                 include_transactions,
-            } => response_channel.send(match height {
-                Number(height) => self.block_queries.by_height(&self.block_memory, height, include_transactions),
-                Finalized | Pending | Latest | Safe => self.block_queries.by_height(&self.block_memory, self.height, include_transactions),
-                Earliest => self.block_queries.by_height(&self.block_memory, 0, include_transactions),
-            }).ok(),
+            } => {
+                let block_height = match height {
+                    Number(height) => height,
+                    Finalized | Pending | Latest | Safe => self.height,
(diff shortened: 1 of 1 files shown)


bulenttastan (approved):
Good catch ✓

birchmd (approved):

moved/src/state_actor.rs
```diff
@@ -192,11 +192,11 @@ impl<
                 height,
                 response_channel,
                 include_transactions,
-            } => response_channel.send(match height {
-                Number(height) => self.block_queries.by_height(&self.block_memory, height, include_transactions),
-                Finalized | Pending | Latest | Safe => self.block_queries.by_height(&self.block_memory, self.height, include_transactions),
-                Earliest => self.block_queries.by_height(&self.block_memory, 0, include_transactions),
-            }).ok(),
+            } => response_channel.send(self.block_queries.by_height(&self.block_memory, match height {
```
bulenttastan, line 195: Just a thought, it might be more readable if we explicitly have the height var:
```rust
let block_height = match height {
    Number(height) => height,
    Finalized | Pending | Latest | Safe => self.height,
    Earliest => 0,
};
response_channel.send(self.block_queries.by_height(&self.block_memory, block_height, include_transactions)).ok()
```

I'll let you decide which version you like.

bulenttastan (approved):
✈️

PR #168, opened 2024-12-04
Feat: op-move integration (part 1)

birchmd:
### Description
My main goal is still to get our integration tests working with op-move responses instead of op-geth. I have not achieved that yet, but I think I am getting close and this PR is big enough so I thought I would push it now. I'll continue iterating once this has been merged.

### Changes
- New command to capture the geth genesis block. This is needed because op-node computes the genesis block hash internally so it will only work if we give it the one it expects.
- Replaced our custom header type with `alloy`'s type
- Replaced merkle root calculation logic with `alloy`'s logic (used for transactions root and receipts root)
- Use `OpTxEnvelope` instead of our envelope type in blocks

### Testing
Existing tests + ongoing integration testing

### Notes
The motivation behind using more `alloy` types and logic is that I have more confidence they are correct than our code is correct.

There is a possibility of some code churn if some of this isn't quite right, but I'd rather a little churn over having to maintain a branch with many changes for a long time.


moved/src/block/root.rs
```diff
@@ -88,108 +88,4 @@ pub struct HeaderForExecution {
...
-        self.transactions_root = transactions_root;
-        self
-    }
-
-    pub fn with_base_fee_per_gas(mut self, base_fee_per_gas: U256) -> Self {
-        self.base_fee_per_gas = base_fee_per_gas;
-        self
-    }
-}
+pub type Header = alloy::consensus::Header;
```
bulenttastan, line 91: We can keep this line higher up.

moved/src/state_actor.rs
```diff
@@ -449,11 +456,10 @@ impl<
             tx_index += 1;
         }
 
-        let receipts_root = receipts
-            .iter()
-            .map(|v| alloy::rlp::encode(&v.receipt))
-            .map(keccak256)
-            .merkle_root();
+        let receipts_root =
+            alloy_trie::root::ordered_trie_root_with_encoder(&receipts, |rx, buf| {
```
bulenttastan, line 460: Can we explain what this does in a short comment? It seems important.

moved/src/types/session_id.rs
```diff
@@ -100,11 +100,6 @@ fn u8_chain_id(genesis_config: &GenesisConfig) -> u8 {
     if genesis_config.chain_id == CHAIN_ID {
         1
     } else {
-        // TODO: Should have some generic fallback algorithm here
-        // (e.g. just take the least significant byte) and a feature
-        // flag to enable it. This would allow people to set their own
-        // custom chain ids. This is not launch-critical since for now
-        // we are only picking chain ids internally.
-        panic!("Unknown chain id: {}", genesis_config.chain_id);
+        genesis_config.chain_id.to_le_bytes()[0]
```
bulenttastan, line 103: nit: I see this U256 → U64 conversion a lot. Any way we can make it a method like `to_u64_lossy()` cause `le` vs `be` could be mishandled?

server/src/lib.rs
```diff
@@ -224,17 +227,72 @@ async fn mirror(
...
         geth_response: &parsed_geth_response,
         op_move_response: &op_move_response,
         port,
     };
     println!("{}", serde_json::to_string_pretty(&log).unwrap());
-    // TODO: use op_move_response
-    let body = hyper::Body::from(geth_response_bytes);
-    Ok(warp::reply::Response::from_parts(geth_response_parts, body))
+
+    // TODO: this is a hack because we currently can't compute the genesis
```
bulenttastan, line 240: We can keep these operations in a separate file, so that this one time operation doesn't take so much code space in here. But it's also Ok to keep it here.

bulenttastan (approved):
Let's ship it ✈️
Left some comments but nothing major.

moved/src/types/session_id.rs
birchmd: This is actually conversion to u8, so the occurrence here is unique in our code. The issue is that Aptos chose u8 to represent chain id for some reason.

moved/src/state_actor.rs
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/168/commits/d8a95d91410e3c960a60dbdead731e42dcc99ff7

server/src/lib.rs
birchmd: Factored out in https://github.com/MovedNetwork/op-move/pull/168/commits/61a1da6b6bc589a7670fad19a9ebab2930bc049d

RomanHodulak (approved):

PR #169, opened 2024-12-05
feat: Add state queries for balance and nonce retrieval of any known block height

RomanHodulak:
### Description
Implements historical lookup of balances and nonces from the blockchain state.

### Changes
- Add state queries for balance and nonce retrieval of any known block height.

### Testing
- Unit test of endpoint
- Unit test of state actor
- Unit test of state queries

### DIFF
--- a/engine-api/src/jsonrpc.rs
+++ b/engine-api/src/jsonrpc.rs
@@ -33,7 +33,7 @@ impl JsonRpcError {
     }
 
     pub fn block_not_found(block_number: BlockNumberOrTag) -> Self {
-        JsonRpcError::without_data(-32001, format!("block not found: {block_number}"))
+        JsonRpcError::without_data(-32001, format!("Block not found: {block_number}"))
     }
 }
 
--- a/engine-api/src/jsonrpc.rs
+++ b/engine-api/src/jsonrpc.rs
@@ -1,4 +1,5 @@
 use {
+    crate::schema::BlockNumberOrTag,
     std::fmt,
     tokio::sync::{mpsc::error::SendError, oneshot::error::RecvError},
 };
@@ -29,6 +30,10 @@ impl JsonRpcError {
 
     pub fn access_state_error<E: fmt::Debug>(e: E) -> JsonRpcError {
         Self::without_data(-1, format!("Failed to access state: {e:?}"))
+    }
+
+    pub fn block_not_found(block_number: BlockNumberOrTag) -> Self {
+        JsonRpcError::without_data(-32001, format!("block not found: {block_number}"))
     }
 }
 
--- a/engine-api/src/methods/get_balance.rs
+++ b/engine-api/src/methods/get_balance.rs
@@ -44,13 +44,13 @@ fn parse_params(request: serde_json::Value) -> Result<(Address, BlockNumberOrTag
 
 async fn inner_execute(
     address: Address,
-    block_number: BlockNumberOrTag,
+    height: BlockNumberOrTag,
     state_channel: mpsc::Sender<StateMessage>,
 ) -> Result<Option<U256>, JsonRpcError> {
     let (tx, rx) = oneshot::channel();
     let msg = Query::BalanceByHeight {
         address,
-        height: block_number,
+        height,
         response_channel: tx,
     }
     .into();
@@ -101,12 +101,8 @@ mod tests {
     #[test_case("pending")]
     #[tokio::test]
     async fn test_execute(block: &str) {
-        let (state_actor, state_channel) = create_state_actor_with_mock_state_queries(
-            1,
-            AccountAddress::new(hex!(
-                "0000000000000000000000000000000000000000000000000000000000000001"
-            )),
-        );
+        let (state_actor, state_channel) =
+            create_state_actor_with_mock_state_queries(AccountAddress::ONE, 1);
 
         let state_handle = state_actor.spawn();
         let request: serde_json::Value = serde_json::json!({
@@ -131,10 +127,10 @@ mod tests {
         let expected_balance = 5;
         let height = 3;
         let (state_actor, state_channel) = create_state_actor_with_mock_state_queries(
-            height,
             AccountAddress::new(hex!(
                 "0000000000000000000000002222222222222223333333333333333333111100"
             )),
+            height,
         );
         let address = "2222222222222223333333333333333333111100";
 
--- a/engine-api/src/methods/get_balance.rs
+++ b/engine-api/src/methods/get_balance.rs
@@ -47,9 +47,9 @@ async fn inner_execute(
     state_channel: mpsc::Sender<StateMessage>,
 ) -> Result<U256, JsonRpcError> {
     let (tx, rx) = oneshot::channel();
-    let msg = Query::GetBalance {
+    let msg = Query::BalanceByHeight {
         address,
-        block_number,
+        height: block_number,
         response_channel: tx,
     }
     .into();
--- a/engine-api/src/methods/get_balance.rs
+++ b/engine-api/src/methods/get_balance.rs
@@ -13,11 +13,12 @@ pub async fn execute(
     state_channel: mpsc::Sender<StateMessage>,
 ) -> Result<serde_json::Value, JsonRpcError> {
     let (address, block_number) = parse_params(request)?;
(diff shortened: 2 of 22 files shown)


moved/src/move_execution/state.rs
```diff
@@ -0,0 +1,360 @@
...
+        self.changes.push(change);
+    }
+
+    pub fn resolver(&self, height: usize) -> impl MoveResolver<PartialVMError> + TableResolver {
+        let mut state = InMemoryStorage::new();
+
+        if let Some(last_index) = self.changes.len().checked_sub(1) {
+            let height = height.min(last_index);
+
+            for change in self.changes[0..=height].iter() {
```
birchmd, line 65: The amount of work here is linear in the height of the request, which will be totally unacceptable after more than a few thousand blocks. I think if our strategy is to store change sets then we also need to store occasional complete state checkpoints to avoid starting from 0 on each query.

I still think the better approach (for now) is to not store change sets at all and instead use the merkle store for history. The idea is that we are already storing the complete set of merkle trie nodes which are part of the merkle tries of all past blocks. To look up a state element at height H we look up the block at H, get its state root R and then index into the merkle trie with root R to find the leaf node we need. This make the amount of work independent of block height. The trade off is that queries will be sensitive to the depth of the trie, but this should grow much slower than H so that it will be fine in the medium term. And avoiding direct merkle queries something we can address in the future with a more clever flat storage implementation.

birchmd (changes requested):

engine-api/src/jsonrpc.rs
```diff
@@ -30,6 +31,10 @@ impl JsonRpcError {
     pub fn access_state_error<E: fmt::Debug>(e: E) -> JsonRpcError {
         Self::without_data(-1, format!("Failed to access state: {e:?}"))
     }
+
+    pub fn block_not_found(block_number: BlockNumberOrTag) -> Self {
+        JsonRpcError::without_data(-32001, format!("block not found: {block_number}"))
```
bulenttastan, line 36: nit: Capitalization of `block` to `Block`

engine-api/src/methods/get_balance.rs
```diff
@@ -13,11 +13,12 @@ pub async fn execute(
...
     let (address, block_number) = parse_params(request)?;
-    let response = inner_execute(address, block_number, state_channel).await?;
+    let response = inner_execute(address, block_number, state_channel)
+        .await?
+        .ok_or(JsonRpcError::block_not_found(block_number))?;
 
     // Format the balance as a hex string
-    Ok(serde_json::to_value(format!("0x{:x}", response))
-        .expect("Must be able to JSON-serialize response"))
+    Ok(serde_json::Value::String(format!("0x{:x}", response)))
```
bulenttastan, line 21: Nice 😎

moved/src/state_actor/queries.rs
```diff
@@ -44,67 +62,83 @@ pub trait StateQueries {
...
-            .into_inner()
+    fn next_version(&mut self) -> Version {
+        self.version += 1;
+        self.version
+    }
+}
+
+/// The `IntoVersionedKv` trait is defined by a single operation [`Self::into_versioned_kv`]. See
+/// its documentation for details.
+trait IntoVersionedKv {
```
bulenttastan, line 116: nit: it should be probably called `IntoVersionedKV` (V is capital) or the long way `IntoVersionedKeyValue`.

moved/src/state_actor/queries.rs
```diff
@@ -149,55 +198,58 @@ impl StateQueries for InMemoryStateQueries {
...
     }
 
     fn nonce_at(
         &self,
         db: &(impl TreeReader<StateKey> + Sync),
         account: AccountAddress,
         height: BlockHeight,
-    ) -> Nonce {
-        let resolver = self.storage.resolver(db, height);
+    ) -> Option<Nonce> {
```
bulenttastan, line 212: Should it be `crate::Result<Nonce>` or `Option<Nonce>`?

moved/src/state_actor/queries.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    fn add(&mut self, changes: ChangeSet) {
+        let version = self.next_version();
+        for (k, v) in changes.into_versioned_kv(version) {
+            self.mem.historic_state.insert(k, v);
+        }
+    }
+
+    fn tag(&mut self) {
+        self.tags.push(self.version);
```
bulenttastan, line 93: Nice, using tag as the block height version 🙌

bulenttastan (commented):
Overall it looks very good. I'll let @birchmd to take a look before approving.

birchmd (commented):
Looks good overall! I left a few additional comments. Rebase onto `main` is also needed to resolve conflicts.

moved/src/state_actor/queries.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    fn balance_at(
+        &self,
+        db: &(impl TreeReader<StateKey> + Sync),
+        account: AccountAddress,
+        height: BlockHeight,
+    ) -> Option<Balance> {
+        let resolver = self.storage.resolver(db, height)?;
+
+        Some(quick_get_eth_balance(&account, &resolver))
```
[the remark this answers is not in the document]
RomanHodulak: I think that is already happening. It only fails on invariant violation (eth_token_invariant_violation).

I have a test for it called `test_query_fetches_zero_balance_for_non_existent_account`

engine-api/src/methods/get_balance.rs
```diff
@@ -45,11 +46,11 @@ async fn inner_execute(
...
     block_number: BlockNumberOrTag,
     state_channel: mpsc::Sender<StateMessage>,
-) -> Result<U256, JsonRpcError> {
+) -> Result<Option<U256>, JsonRpcError> {
     let (tx, rx) = oneshot::channel();
-    let msg = Query::GetBalance {
+    let msg = Query::BalanceByHeight {
         address,
-        block_number,
+        height: block_number,
```
bulenttastan, line 53: We should just rename the `inner_execute` input to `height: BlockNumberOrTag`. Same applies to get_nonce.

engine-api/src/methods/get_nonce.rs
```diff
@@ -100,7 +104,12 @@ mod tests {
     #[test_case("pending")]
     #[tokio::test]
     async fn test_execute(block: &str) {
-        let (state_actor, state_channel) = create_state_actor();
+        let (state_actor, state_channel) = create_state_actor_with_mock_state_queries(
+            1,
+            AccountAddress::new(hex!(
+                "0000000000000000000000000000000000000000000000000000000000000001"
```
bulenttastan, line 110: We could simply use `AccountAddress::ONE`

engine-api/src/methods/mod.rs
```diff
@@ -145,4 +153,46 @@ pub mod tests {
         channel.send(msg).await.map_err(access_state_error).unwrap();
         receiver.await.map_err(access_state_error).unwrap();
     }
+
+    pub fn create_state_actor_with_mock_state_queries(
+        height: u64,
+        address: AccountAddress,
```
bulenttastan, line 159: Feels like `address` should be the first param, then comes the `height`.

moved/src/state_actor/queries.rs
```diff
@@ -0,0 +1,672 @@
...
+        storage.add(inc_one_nonce(2, &mut state, addr));
+        storage.tag();
+
+        let query = InMemoryStateQueries::new(storage);
+
+        let actual_nonce = query
+            .nonce_at(state.db(), addr, 1)
+            .expect("Block height should exist");
+        let expected_nonce = 1u64;
+
```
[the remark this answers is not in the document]
RomanHodulak: There is another test for it, this one checks older balance.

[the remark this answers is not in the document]
RomanHodulak: Oh now I see it, you're right.

bulenttastan (approved):
Looks good. This was a big PR 😮‍💨

birchmd (approved):
Approving to unblock. Historical lookups are an important feature! I still think the new `OnTx` and `OnTxBatch` might be unnecessary complexity, but we can always iterate on the design later.

PR #171, opened 2024-12-06
Feat(rpc): Implement eth_call by transaction simulation

bulenttastan:
### Description
`eth_call` implementation by simulating an entry function or script transaction. Part of #47.

### Changes
- `call_transaction` method to execute entry fn or script input
- `deploy_contract` helper test method to set up `eth_call` test
- Tests for entry fn and script bytecode inputs

### Testing
✓ New tests for entry fn and script

### Notes
Refactored `simulate_transaction` into `simulate.rs` under `move_execution`.

### DIFF
--- a/moved/src/move_execution/simulate.rs
+++ b/moved/src/move_execution/simulate.rs
@@ -1,0 +1,99 @@
+use {
+    crate::{
+        block::HeaderForExecution,
+        genesis::config::GenesisConfig,
+        move_execution::{
+            create_move_vm, create_vm_session, execute_transaction, quick_get_nonce,
+            BaseTokenAccounts,
+        },
+        primitives::{ToMoveAddress, B256},
+        types::{
+            session_id::SessionId,
+            transactions::{
+                NormalizedEthTransaction, NormalizedExtendedTxEnvelope, ScriptOrModule,
+                TransactionData, TransactionExecutionOutcome,
+            },
+        },
+        Error::InvalidTransaction,
+        InvalidTransactionCause,
+    },
+    alloy::rpc::types::TransactionRequest,
+    move_binary_format::errors::PartialVMError,
+    move_core_types::resolver::MoveResolver,
+    move_table_extension::TableResolver,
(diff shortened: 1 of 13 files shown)


moved/src/move_execution/simulate.rs
```diff
@@ -0,0 +1,99 @@
...
+        &B256::random(),
+        state,
+        genesis_config,
+        0,
+        base_token,
+        block_header,
+    )
+}
+
+pub fn call_transaction(
```
birchmd, line 61: This function currently skips the nonce check which is present in `execute_transaction`. We could add it in, but that would be duplicating some code. And in the future when there are more pre-/post-transaction operations (e.g. charging and refunding gas) then those operations would need to be duplicated here as well.

I understand the point of having this function separate is to enable getting return values for view calls, but I think some refactoring will be necessary to have this function reuse logic from `execute_transaction`. Probably the easiest way to do that is to factor out the pre-transaction check as another function which we can use here as well as in `execute_transaction`. Similarly, we need to make sure this function gets the gas charging functionality when it gets implemented in the future.

birchmd (commented):
This looks pretty good overall! I just have a few comments/questions.

moved/src/move_execution/simulate.rs
```diff
@@ -0,0 +1,99 @@
...
+    request: TransactionRequest,
+    state: &(impl MoveResolver<PartialVMError> + TableResolver),
+) -> crate::Result<Vec<u8>> {
+    let tx = NormalizedEthTransaction::from(request.clone());
```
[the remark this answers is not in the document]
bulenttastan: It fails without `.clone()`, complains that request doesn't derive `Copy`. I change it to construct `tx` after getting the nonce.

moved/src/move_execution/simulate.rs
bulenttastan: Working on this.
I'll separate the transaction metadata checks from `canonical` and use it in both places.

birchmd (approved):

PR #172, opened 2024-12-11
Fix: Include withdrawals root in block header

birchmd:
### Description
Adds the `withdrawals_root` to the block header. This is required for `op-batcher` to work properly with `op-move`.

### Changes
- Use `alloy` to compute withdrawals root

### Testing
- Continued integration testing work

### Notes
I'm not sure if this implementation is totally right (the withdrawals might need a different serialization than the default for example), but it does work when the withdrawals array is empty, which is the only case we consider for now. We can continue iterating on this when we actually have withdrawals.


bulenttastan (approved):
Seems correct to me, based on:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4895.md#execution-payload-validity

PR #173, opened 2024-12-13
Feat(transactions): correct deposit flow using OP L2 contracts

birchmd:
### Description
This PR changes the way we process deposit-type transactions to match the usual OP stack deposit flow. With this change the `use_optimism_bridge` part of the end-to-end integration test now passes when RPC responses come from `op-move` instead of `op-geth`! 🚀 

### Changes
- New private Move function in EVM module to enable the system to call the EVM native directly.
- When processing a deposit-type transaction, execute the EVM

### Testing
Update to `test_deposit_tx` which now uses an example deposit transaction taken from our end-to-end test (so it actually came from `op-node`.

### Notes
I kept the option for test code to mint ETH without going through the EVM to keep it easy to fund accounts when writing unit tests.


genesis-builder/framework/evm/sources/Evm.move
```diff
@@ -117,6 +117,12 @@ module Evm::evm {
         amount
     }
 
+    // A private function used by the system to call the EVM native.
+    // (For some reason we cannot call the native function directly)
+    fun system_evm_call(caller: address, to: address, value: u256, data: vector<u8>): EvmResult {
```
bulenttastan, line 122: Do we need the owner check just for extra security?

moved/src/move_execution/deposited.rs
```diff
@@ -1,31 +1,144 @@
...
+        .expect("EVM changes must merge with other session changes");
+
+    Ok(TransactionExecutionOutcome::new(
+        vm_outcome, changes, gas_used, logs, None,
+    ))
+}
+
+/// This function is only used in tests.
+/// It allows us to mint ETH directly without going through the EVM.
+#[cfg(any(feature = "test-doubles", test))]
```
bulenttastan, line 132: We probably need to remove `"test-doubles"` feature then.

moved/src/move_execution/deposited.rs
```diff
@@ -1,31 +1,144 @@
...
+));
+
 pub(super) fn execute_deposited_transaction(
     tx: &DepositedTx,
     tx_hash: &B256,
     state: &(impl MoveResolver<PartialVMError> + TableResolver),
     genesis_config: &GenesisConfig,
     block_header: HeaderForExecution,
 ) -> crate::Result<TransactionExecutionOutcome> {
+    #[cfg(any(feature = "test-doubles", test))]
```
bulenttastan, line 43: This `#cfg` isn't needed I assume.

bulenttastan, line 43: Is this safe to not cause any unwanted deposit injections?

moved/src/move_execution/deposited.rs
```diff
@@ -1,31 +1,144 @@
...
+    if tx.data.is_empty() {
```
bulenttastan, line 44: Is this safe to not cause any unwanted deposit injections?

moved/src/move_execution/deposited.rs
```diff
@@ -1,31 +1,144 @@
...
+        if !evm_outcome.is_success {
+            return Err(UserError::DepositFailure(evm_outcome.output));
+        }
+        // TODO: Should consider ERC-20 deposits here too?
+        let bridge_log = evm_outcome
+            .logs
+            .iter()
+            .find(|l| l.topics()[0] == ETH_BRIDGE_FINALIZED)
+            .ok_or(UserError::DepositFailure(evm_outcome.output))?;
+        let dest_address = AccountAddress::new(bridge_log.topics()[2].0);
```
bulenttastan, line 99: nit: Short explanation of event topic structure would be helpful here.

genesis-builder/framework/evm/sources/Evm.move
birchmd: If it's a private function there is no way for an external caller to invoke it right? I don't think we also need to check the caller.

moved/src/move_execution/deposited.rs
birchmd: This code literally does not exist outside of tests/test-doubles (this is how the conditional compilation works) so it should be safe.

bulenttastan: True. test-double was to allow testing between crates, not to open to production 👍

moved/src/types/transactions.rs
```diff
@@ -456,7 +476,6 @@ mod tests {
         rlp_roundtrip(&Bytes::from_static(&hex!("02f86f82a45580808346a8928252089465d08a056c17ae13370565b04cf77d2afa1cb9fa8806f05b59d3b2000080c080a0dd50efde9a4d2f01f5248e1a983165c8cfa5f193b07b4b094f4078ad4717c1e4a017db1be1e8751b09e033bcffca982d0fe4919ff6b8594654e06647dee9292750")));
     }
 
-    #[cfg(test)]
```
RomanHodulak, line 459: Should this be kept?

genesis-builder/framework/evm/sources/Evm.move
bulenttastan: As long as we don't skip visibility in the execution, this should be safe.

bulenttastan (approved):
The code looks really good. Good to have the deposit functionality.

moved/src/move_execution/deposited.rs
RomanHodulak: I think this can cause unwanted side effects - what if I want to test the normal flow?

RomanHodulak (approved):
Great to have this working.

moved/src/types/transactions.rs
birchmd: It was redundant because it appears in a block that is already marked as `#[cfg(test)]`.

moved/src/move_execution/deposited.rs
RomanHodulak: Thanks for providing the rationale. It's good to be aware that we make it impossible to test one edge case.

I'd say the decision to alter a function's behavior should be made explicit from the call site e.g. by the use of a test only dependency. As the caller, I should tell the function what to do, not adapt to how the function behaves.

Not blocking it though, you got my approve.

moved/src/move_execution/deposited.rs
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/173/commits/8f87282c09a8ad513ea7f833fc8f72b29dc0f926

moved/src/move_execution/deposited.rs
birchmd: As discussed in the meeting today, we will leave the PR as-is for now. While we all agree this the current state is sub-optimal, modifying the tests to use the proper deposit flow can be done in a future PR (task for post-devnet launch).

PR #179, opened 2024-12-24
feat(gas): add l2 gas cost calculation

### LINKED ISSUE #160: Charge gas for computation on L2

ilchu:
### Description

Closes #160.

### Changes

- Pass through L2 costs when executing canonical txs
- Expand the `L1GasFee` trait with L2 functionality
- Charging and refunding the costs before and after execution respectively

### Testing

Will add tests in the spirit of what we have for L1 as @bulenttastan mentioned in Slack.

L1 and L2 costs will have to be coupled together due to nature of execution, so I see appropriate test cases as such:
* Charging L1 + L2 costs on success / failure correctly
* Proper L2 refunds on success / failure
* L2 costs never exceeding `gas_limit * effective_price`

### Notes

Things left to do:
- [x] having a common `GasFee` trait seems to be an over-generalization (especially due to the point from the next check mark), and I will rework it to be separate for L1/L2
- [x] do not include blob fees in calculation and have separate custom base fee for L2
- [x] refunding flow after execution
- [x] unit tests basics
- [x] implement and fix cases from the testing section
- [x] Record actual gas used in tx receipts


moved/src/move_execution/tests/l2_cost.rs
```diff
@@ -0,0 +1,49 @@
+use super::*;
+
+#[test]
+fn test_treasury_charges_l1_cost_to_sender_account_on_success() {
```
bulenttastan, line 4: The test name still has `l1` in it instead of `l2`.

Also I think we should keep the `L1` and `L2` cost tests together, no need for separate files here.
We can call the new file `gas.rs` or `gas_cost.rs` (leaving the choice to you) and keep everything together.

moved/src/move_execution/eth_token.rs
```diff
@@ -45,6 +45,15 @@ pub trait BaseTokenAccounts {
         gas_meter: &mut G,
     ) -> Result<(), crate::Error>;
 
+    fn refund_l2_cost<G: GasMeter>(
```
bulenttastan, line 48: Should we just name this method to just `refund` 🤔?
Here's my thought, from the `BaseTokenAccounts` perspective it would just handle gas costs and refunds. However the layer above will determine the business logic of charging or refunding for L1 or L2. This might be wrong, but it's a thought, since we also change the charging to generic `charge_gas_cost`.

ilchu: Sounds like it would be a clean solution indeed, but my only concern would be if we ever need refunds on L1? AFAICT at least for deposited transactions this explicitly never happens.

moved/src/move_execution/canonical.rs
```diff
@@ -153,13 +166,21 @@ pub(super) fn execute_canonical_transaction(
         }
     };
 
+    let gas_used = total_gas_used(&gas_meter, genesis_config);
+    base_token.refund_gas_cost(
```
bulenttastan, line 170: Refunding cause test failures. I think there is a logical issue here. We compute the `gas_used` but we haven't charged it yet, so refunding it before charging seems wrong.

moved/src/move_execution/tests/utils.rs
```diff
@@ -150,7 +155,7 @@ impl TestContext {
         let treasury_address = AccountAddress::ONE;
         let base_token = MovedBaseTokenAccounts::new(treasury_address);
         let mut transaction = TestTransaction::new(tx, tx_hash);
-        transaction.with_cost(l1_cost, base_token);
+        transaction.with_cost(l1_cost, l2_cost, base_token);
```
bulenttastan, line 158: The assertion below should be for the sum of L1 and L2 costs, since the cost can come from either one now:
```rust
assert_eq!(treasury_balance, U256::from(l1_cost + l2_cost));
```

moved/src/move_execution/canonical.rs
ilchu: It's actually due to me setting it to zero, and it was properly charged up to a limit [here](https://github.com/MovedNetwork/op-move//blob/feat/ilchu/l2_gas_calc/moved/src/move_execution/canonical.rs#L74-L80). The issue with the tests was due to L2 costs being set by me to 0 in tests, whereas actual gas meter did accrue some units, and it was failing trying to refund back the sum that wasn't properly deposited into treasury in the first place.

moved/src/move_execution/tests/utils.rs
ilchu: Good point! But actually I need to make it even smarter, as that wouldn't account for refunds. So I'm currently considering what is the most ergonomic way to refactor this assert.

moved/src/move_execution/canonical.rs
```diff
@@ -153,13 +166,21 @@ pub(super) fn execute_canonical_transaction(
...
+        &sender_move_address,
+        gas_used - l2_cost,
```
birchmd, line 172: I think you mentioned this in the meeting, but I wanted to confirm here that you do need to pass along the price because you need to convert the gas amount into a token amount by multiplying by the price. With the price set to zero then there would be no amount charged or refunded so then the existing tests will pass.

moved/src/move_execution/eth_token.rs
```diff
@@ -87,6 +96,26 @@ impl BaseTokenAccounts for MovedBaseTokenAccounts {
         )
     }
 
+    fn refund_gas_cost<G: GasMeter>(
+        &self,
+        to: &AccountAddress,
+        amount: u64,
+        session: &mut Session,
+        traversal_context: &mut TraversalContext,
+        gas_meter: &mut G,
```
bulenttastan, line 105: I think it can create the `free_gas_meter` internally inside the function, no need to take in as an input.

go-ethereum
bulenttastan, line 1: You didn't mean to add this and the `op-geth` as submodules right?
If you did, they should probably go under `server/src/tests/`

moved/src/state_actor.rs
```diff
@@ -427,12 +435,20 @@ impl<
...
 
         // TODO: parallel transaction processing?
         for (tx_hash, tx, l1_cost_input) in transactions {
-            let Ok(normalized_tx) = tx.clone().try_into() else {
+            let Ok(normalized_tx): Result<NormalizedExtendedTxEnvelope, _> = tx.clone().try_into()
+            else {
                 continue;
             };
+            // TODO: implement gas limits etc. for `ExtendedTxEnvelope` so that
+            // l2 gas inputs can be constructed at an earlier stage and stored in mempool?
```
bulenttastan, line 447: nit: no need for question mark ?

moved/src/move_execution/tests/transfer.rs
```diff
@@ -1,3 +1,5 @@
+use std::u64;
```
bulenttastan, line 1: nit: you can check clippy locally:
`cargo clippy --all-features --all-targets`

moved/src/move_execution/tests/gas_cost.rs
```diff
@@ -0,0 +1,150 @@
...
+    // Set a gas limit lower than the cost of operation, but still enough to pay L1 costs
+    let l2_gas_limit = L1_GAS_COST;
+    let outcome = ctx.transfer(
+        receiver,
+        transfer_amount,
+        l1_cost,
+        l2_gas_limit,
+        l2_gas_price,
+    );
+    assert!(matches!(
```
bulenttastan, line 103: Using `matches!()` is cool 🙌

bulenttastan (commented):
Looks really good, solid work 👏
Left some minor comments and leaving it to @birchmd to go over too.

moved/src/move_execution/canonical.rs
```diff
@@ -82,12 +95,15 @@ pub(super) fn verify_transaction(
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)]
```
bulenttastan, line 98: Let's create a `P2` issue to fix the too many args issue. We'll probably create a few structs for gas and execution context to wrap the inputs together.

moved/src/move_execution/gas.rs
```diff
@@ -82,15 +90,41 @@ impl<T: AsRef<[u8]>> From<T> for L1GasFeeInput {
...
+impl L2GasFeeInput {
+    pub fn new(gas_limit: u64, effective_gas_price: U256) -> Self {
+        Self {
+            gas_limit,
+            effective_gas_price,
+        }
+    }
+}
+
+impl From<(u64, U256)> for L2GasFeeInput {
```
bulenttastan, line 110: Nice tuple to struct shortcut 🙌

moved/src/move_execution/tests/transfer.rs
ilchu: I did a few times when cleaning up, but it's my nvim completion adding unneeded imports without me asking. Thanks for pointing out, still.

go-ethereum
ilchu: Not intended, happened accidentally.

moved/src/move_execution/canonical.rs
bulenttastan: Created a new issue for it: https://github.com/MovedNetwork/op-move/issues/184

moved/src/move_execution/eth_token.rs
ilchu: That's definitely more sleek, thanks for the suggestion! Did see it in the codebase, but forgot by the point of writing the code.

moved/src/move_execution/tests/transfer.rs
```diff
@@ -111,13 +115,16 @@ fn test_eoa_base_token_transfer() {
     // Should fail when transfer is larger than account balance
     let receiver = ALT_EVM_ADDRESS;
     let transfer_amount = mint_amount.saturating_add(U256::from(1));
-    let outcome = ctx.transfer(receiver, transfer_amount, 0);
-    outcome.vm_outcome.unwrap_err();
+    // Still need to set gas limit for proper functioning of the gas meter
+    let outcome = ctx.transfer(receiver, transfer_amount, 0, 100, U256::ZERO);
+    // Insufficient intrinsic gas
```
birchmd, line 120: What is this comment about? In this test it looks like the error should be insufficient balance, not anything about gas.

moved/src/state_actor.rs
```diff
@@ -427,12 +435,20 @@ impl<
             .peek()
             .and_then(|(_, v, _)| v.as_deposited())
             .map(|tx| self.l1_fee.for_deposit(tx.data.as_ref()));
+        let l2_fee = self.l2_fee.with_gas_fee_multiplier(U256::from(1));
```
birchmd, line 438: Seems like a constant should be used here instead of a hardcoded literal. I think we maybe have a multiplier constant defined elsewhere already?

birchmd (commented):
Nice work! I've only left a couple minor comments.

moved/src/move_execution/tests/transfer.rs
ilchu: I guess I left it when debugging the broken test and forgot to revert it, good catch!

moved/src/state_actor.rs
ilchu: We did have constants specific for Ecotone fees or block EIP-1559 fees, but I believe it would be a bit confusing to reuse any of them here. I added a const to the `CreateL2GasFee` in a similar fashion and a helper method. Hope it makes things more explicit.

birchmd (approved):

moved/src/state_actor.rs
birchmd: Yeah, looks good. Thanks!

PR #181, opened 2024-12-30
Feat(L2): Generic L2 contract request handler

### LINKED ISSUE #98: Design: withdrawals from the L2
Make sure that we can withdraw based tokens from the L2 to the L1. Can probably use Optimism's design more or less directly.

bulenttastan:
### Description
All the L2 contract requests go through this canonical transaction execution, including the withdrawal requests. This started as a way to initiate withdrawals, however it extended to handle every L2 request. Part of #98

### Changes
- Handle L2 call requests in canonical transaction execution
- Burn any payable ETH value and then execute the L2 contract through `evm` Move native extension
- Implement `burn_eth` method in `eth_token`
- Return the EVM logs as transaction outcome

### Testing
✓ Test to initiate a withdrawal request

### DIFF
--- a/moved/src/error.rs
+++ b/moved/src/error.rs
@@ -72,6 +72,8 @@ pub enum UserError {
     InvalidSignature(#[from] alloy::primitives::SignatureError),
     #[error("Error during EVM execution for L2 bridge {0:?}")]
     DepositFailure(Vec<u8>),
+    #[error("L2 contract call failure")]
+    L2ContractCallFailure,
 }
 
 /// The error caused by invalid transaction input parameter.
--- a/moved/src/move_execution/canonical.rs
+++ b/moved/src/move_execution/canonical.rs
@@ -6,7 +6,7 @@ use {
             create_move_vm, create_vm_session,
             eth_token::{BaseTokenAccounts, TransferArgs},
             evm_native,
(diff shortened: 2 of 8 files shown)


birchmd (approved):
This is a neat idea, nice work!

This PR provides a shortcut for calling the L2 contracts without going through the MoveVM. This works nicely when the transaction only involves the L2 contracts (as is the case with EOA withdraw), though I still think it is good we have Move interfaces for interacting with the L2 contracts for more general use cases (e.g. triggering a withdraw from a Move-based escrow contract).

The same functionality this PR provides could be done via a Move script invoking the `initiateWithdrawal()` function on the Move interface for the `L2toL1MessageParser`, but I don't think that having a more convenient way to accomplish this common flow is a bad thing.

However, there is one notable difference between this flow and the Move script flow. In this flow the ETH is burned while in the script flow the ETH would be [transferred to the EVM Move address](https://github.com/MovedNetwork/op-move/blob/main/genesis-builder/framework/evm/sources/Evm.move#L125). I think it is important for us to ensure that two way of accomplishing the same thing have the same state changes. I think the best way to resolve the current discrepancy is to give the EVM account the burn capability during genesis and have the Move contract burn the tokens just as they are burned here. This change can be done in a future PR.

PR #182, opened 2024-12-30
Feat(server): Use op-move response instead of op-geth

### LINKED ISSUE #47: Return op-move response

birchmd:
### Description
Closes #47 

### Changes
- Server now uses op-move response instead of op-geth (op-geth is still present for getting genesis hash)
- Bug fix in `eth_feeHistroy` param order
- Increase value returned by `eth_estimateGas` because it was not accounting for the L1 fee transfer (a proper solution to this problem will be needed in the future)
- It is an invalid transaction error if the base fee fails to charge, as opposed to another kind of error
- Integration test logs each OP stack program to a separate log file
- Integration test now deploys a simple Move contract instead of an ERC-20 EVM contract

### Testing
- Integration test


moved/src/state_actor.rs
```diff
@@ -247,9 +247,10 @@ impl<
                     Finalized | Pending | Latest | Safe => self.height,
                     Earliest => 0,
                 };
+                // TODO: simulation should account for gas from non-zero L1 fee
                 let outcome = simulate_transaction(transaction, self.state.resolver(), &self.genesis_config, &self.base_token, block_height);
                 match outcome {
-                    Ok(outcome) => response_channel.send(Ok(outcome.gas_used)).ok(),
+                    Ok(outcome) => response_channel.send(Ok(1000 * outcome.gas_used)).ok(),
```
bulenttastan, line 253: We can make the 1000 multiplier a constant or leave a comment why it's 1000.

server/src/tests/integration.rs
```diff
@@ -275,6 +276,7 @@ fn init_and_start_op_geth() -> Result<Child> {
         .output()
         .context("Call to state dump failed")?;
     check_output(output);
+    let op_geth_logs = std::fs::File::create("op_geth.log").unwrap();
```
bulenttastan, line 279: nit: We can just say `File::create` instead of `std::fs::File::create` since `File` is imported.

bulenttastan (approved):
IT WORKS ✨

It might be worth to stop mirroring `op-geth` after getting the genesis hash.

moved/src/state_actor.rs
birchmd: The multiplication here isn't correct, it's just needed to make the test pass because the current gas estimation doesn't include the cost of doing the transfer to pay for the L1 fee. The multiplier must be removed once the simulation logic is improved, per the TODO above.

bulenttastan: I see that's what the TODO was about. Makes sense

server/src/tests/integration.rs
birchmd: Fixed in https://github.com/MovedNetwork/op-move/pull/182/commits/570a99c128cf5d5d4950ed6c5ecab3167dd5c2e7

moved/src/move_execution/canonical.rs
```diff
@@ -61,13 +61,15 @@ pub(super) fn verify_transaction(
...
-    )?;
+    base_token
+        .charge_l1_cost(
+            &sender_move_address,
+            l1_cost,
+            session,
+            traversal_context,
+            gas_meter,
+        )
+        .map_err(|_| InvalidTransaction(InvalidTransactionCause::FailedToPayL1Fee))?;
```
ilchu, line 72: Nice! I was just figuring out what to do in case of L2 gas limit overflows. Guess I'll follow the same pattern.

server/src/tests/integration.rs
```diff
@@ -76,8 +77,8 @@ async fn test_on_ethereum() -> Result<()> {
     // 10. Test out the OP bridge
     use_optimism_bridge().await?;
 
-    // 11. Test out an ERC20 token
-    deploy_erc20_token().await?;
+    // 11. Test out a simple Move contract
+    deploy_move_counter().await?;
```
ilchu, line 81: So enabling Move contract deployments like this is the biggest thing that was enabled by shift to op-move AFAICT. Does it also mean that with it the ERC20 functionality is gone? Or is it just a replacement of a less complex test case with a more complex one?

bulenttastan: ERC20 would be replaced with Move fungible tokens. However, in the future I think we will support ERC20 exactly as is to be compatible with ERC20 and ERC721 bridges between L1 and L2.

PR #183, opened 2025-01-02
Feat(storage): Switch Merkle implementation to `jmt` library 

birchmd:
### Description
The goal of this PR is to make storage proofs of our state verifiable on EVM smart contracts by using the Keccak256 hash function in our Merkle trie (which is natively supported on EVM). The Aptos library was not flexible in its hash function choice, so we have switched to a different library which does allow us to choose our hash function.

### Changes
- Replace Aptos Merkle library with `jmt`
- Delete most code related to historical value storage because the `jmt` library does not have the same indirection as the Aptos library had (see also the note below).

### Testing
- Existing tests

### Notes
In the `jmt` library, the values are directly returned from the `TreeReader` instead of having the key + version indirection that was present in the aptos implementation. This means the `StateMemory` struct is now less useful than it used to be. I left it as it is because I assume there will be a large refactor of the storage implementation when we move to a persistent storage layer anyway. But I just wanted to note here that we do not necessarily want to mimic the current architecture of the in-memory storage when we write the persistent one.


bulenttastan (approved):
Nice. Seems like `jmt` has slightly better interface for historic state access.

PR #186, opened 2025-01-08
Feat(rpc): eth_getProof implementation

birchmd:
### Description
Adds an implementation of `eth_getProof` to our RPC. The proofs returned are _not_ Ethereum MPT proofs, rather they are JMT proofs because we use a different Merkle tree implementation.

### Changes
- `eth_getProof` rpc method
- New method on `StateQueries` trait for the `eth_getProof` method

### Testing
- Basic sanity test in `get_proof.rs`
- Integration/devnet testing

### Notes
I needed to redefine some types from the `jmt` library because they have private fields that I need access to when constructing the output for `eth_getProof`.

### DIFF
--- a/engine-api/src/methods/get_proof.rs
+++ b/engine-api/src/methods/get_proof.rs
@@ -1,0 +1,157 @@
+use {
+    crate::{
+        json_utils::{self, access_state_error},
+        jsonrpc::JsonRpcError,
+    },
+    alloy::{
+        eips::BlockNumberOrTag,
+        primitives::{Address, U256},
+    },
+    moved::types::{
+        queries::ProofResponse,
+        state::{Query, StateMessage},
+    },
+    tokio::sync::{mpsc, oneshot},
+};
+
+pub async fn execute(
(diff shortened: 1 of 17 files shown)


bulenttastan (dismissed):

engine-api/src/methods/get_proof.rs
```diff
@@ -0,0 +1,159 @@
...
+        primitives::{Address, U256},
+    },
+    moved::types::{
+        queries::ProofResponse,
+        state::{Query, StateMessage},
+    },
+    tokio::sync::{mpsc, oneshot},
+};
+
+pub async fn execute(
```
bulenttastan, line 17: The `execute` method is missing in `request.rs` and the naming is missing in `method_name.rs` to redirect the client calls.

birchmd: Good catch, thanks. Fixed in https://github.com/MovedNetwork/op-move/pull/186/commits/e3c1a30c897f53053ae7bff1496c3cfd9b738d6e

PR #187, opened 2025-01-13
feat(state): Replace JMT with MPT

birchmd:
### Description
To accommodate Ethereum tooling that relies on Merkle Patricia Trie (MPT) proofs, we replace the Jellyfish Merkle Tree (JMT) library with `eth_trie` instead. This should make `eth_getProof` work properly with existing software that processes Ethereum Merkle proofs.

### Changes
- Remove `jmt` and add `eth_trie`
- State queries are now versioned by state root hash instead of version numbers, this means we no longer take any action on each transaction processing (`on_tx_in_memory` is a noop).

### Testing
- Existing tests

### Notes
- In a future PR we may want to add more robust tests, e.g. specifically deleting data from the state and ensuring it is still accessible by historical queries.


moved/src/state_actor.rs
```diff
@@ -650,11 +650,17 @@ impl<
     > StateActor<S, P, H, R, G, L1G, L2G, B, Q, M, InMemoryStateQueries>
 {
     pub fn on_tx_in_memory() -> OnTx<Self> {
-        Box::new(|| Box::new(|state, changes| state.state_queries.add(changes)))
+        Box::new(|| Box::new(|_state, _changes| ()))
```
bulenttastan, line 653: Is this method not used any more? just returning `()`

moved/src/state_actor.rs
```diff
@@ -650,11 +650,17 @@ impl<
...
     }
 
     pub fn on_tx_batch_in_memory() -> OnTxBatch<Self> {
-        Box::new(|| Box::new(|state| state.state_queries.tag()))
+        Box::new(|| {
```
bulenttastan, line 657: nit: The curly brackets here and below seem not needed.

moved/src/storage.rs
```diff
@@ -97,62 +102,48 @@ impl State for InMemoryState {
...
-            .write_tree_update_batch(tree_update_batch)
-            .expect("Fails on duplicate key or storage write. In memory storage cannot fail.");
 
-        root_hash.to_h256()
+        let mut trie = self.tree();
+        for (k, v) in values {
+            let key_bytes = k.key_hash();
+            let value_bytes = v
+                .as_ref()
+                .map(|x| bcs::to_bytes(x).expect("Value must serialize"));
```
bulenttastan, line 127: Is this the reason why the proofs don't come in as the Proposer expects?
As in, the tree keeps the `change_set` instead of an `account` info.

bulenttastan:
This change looks good to me as switching from JMT to MPT. The underlying state root calculation might be incorrect with what the OP Stack expects, but we can focus on that in another PR.
Also having another eye by @RomanHodulak would be good.

RomanHodulak (approved):
I wonder how hard it would be to make the code generic over which trie implementation is used.

moved/src/state_actor.rs
birchmd: No longer used, correct. But I am hesitant to remove it in case we end up needing it later. We are still iterating on this after all. Once we have something stable, if this is no longer used then we can remove it entirely.

moved/src/state_actor.rs
birchmd: I think the formatter put them there automatically because the closures span multiple lines.

moved/src/storage.rs
birchmd: Perhaps this is the reason. Depends on if the value hash is included in the proof or not. We'll have to look into it more.

birchmd:
> generic over which trie implementation is used

I suppose we'd need to write our own interface (trait) and then implement it for specific tries. This probably wouldn't be a huge effort, but I also think probably not worth while since we don't want to change the trie implementation on a regular basis. We want to get the trie working and then never touch it again.

moved/src/state_actor.rs
bulenttastan: That's valid, we should keep it then ✓

birchmd:
With two approvals, I'll go ahead and merge this. We'll keep iterating on it as we debug why `op-proposer` is still unhappy with the merkle proofs we provide.

PR #188, opened 2025-01-14
Fix(trie): store EVM account information under the expected key

birchmd:
### Description
To verify a Merkle proof, you need to know the expected state root and which key it is supposed to be a proof for. Ethereum tooling expects the key corresponding to the proof of an account to be `keccack256(address)`. Previously, we were following the Aptos convention of using `StateKey::resource`, however this PR changes to the Ethereum convention for EVM account resources so that Ethereum tooling can verify the proofs automatically.

### Changes
- Move resources corresponding to EVM accounts are stored in the trie under the key `keccack256(address)`

### Testing
- Existing tests
- New integration test showing the proofs from `eth_getProof` can be verified successfully.

### Notes
- There is a new `TODO` because the obvious corollary of this change is that the EVM account storage should be handled differently as well for the storage proofs to be verified. This is a little trickier to handle so I'm leaving the `TODO` for a future PR.
- The new test sets out a pattern for how we could do other integration tests in the `server` crate by directly sending JSON payloads to the state actor. I didn't make the code very reusable in this PR (just the function for creating the state actor), but this is something we will likely want to do in the future when we write more integration tests.


RomanHodulak (approved):
So normally, the EVM native extension stores accounts as custom resource type with `Account_` prefix. But with this change, internally, the resources inside EVM native Move module are saved such that their 20-byte EVM address is used as their key in the state trie, instead of the standard fully qualified resource key.

The motivation for this change is to exactly match what would be in a standard EVM trie to satisfy proof generated from the client side which assumes that the EVM addresses are used as keys in the trie.

Does that mean that the proofs are only requested for resources inside EVM native Move module?

birchmd:
> Does that mean that the proofs are only requested for resources inside EVM native Move module?

We currently only support proofs for resources inside the EVM, yes. This is because we currently only need this endpoint to prove the state of the L2 contracts we are using from the OP-stack. We will need to think about what merkle proofs we want to support for the MoveVM itself. I think this depends on the needs of Move developers we on-board to the platform. They might prefer a different endpoint than `eth_getProof` because they are using tooling from the Move ecosystem instead of EVM tooling.

PR #189, opened 2025-01-15
feat: Integrate Moved with RocksDB persistent storage engine to back the blockchain state trie

RomanHodulak:
### Description
Adds a crate that integrates `moved` with `rocksdb`.

### Changes
- Added a crate that integrates `moved` with `rocksdb`.

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt


rocksdb/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+mod r#move;
```
bulenttastan, line 1: Maybe we can call this `persistence` or just `persist` as in the storage layer.

rocksdb/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "moved-rocksdb"
```
bulenttastan, line 2: We can probably call this `moved-storage` with folder name `storage`.

rocksdb/src/move.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    fn insert(&self, key: &[u8], value: Vec<u8>) -> Result<(), Self::Error> {
+        self.db().put(key, value)
+    }
+
+    fn remove(&self, key: &[u8]) -> Result<(), Self::Error> {
+        self.db().delete(key)
+    }
+
+    fn flush(&self) -> Result<(), Self::Error> {
```
bulenttastan, line 97: Will `flush` be used, like will we force Rocks DB to write to files after a block is done?

rocksdb/src/move.rs
```diff
@@ -0,0 +1,100 @@
...
+        let db = self.db();
+
+        db.get(key).map(|v| v.map(|v| Bytes::copy_from_slice(&v)))
+    }
+}
+
+impl<'db> ResourceResolver for RocksEthTrieDb<'db> {
+    type Error = rocksdb::Error;
+
+    fn get_resource_bytes_with_metadata_and_layout(
```
bulenttastan, line 54: Should we make this fn name less explicit like call it only `get_resource`?

rocksdb/src/moved.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl<'db> RocksDbState<'db> {
+    const STATE_ROOT_KEY: &'static str = "state_root";
+
+    pub fn new(db: Arc<RocksEthTrieDb<'db>>) -> Self {
+        let state_root = db
+            .db()
+            .get(Self::STATE_ROOT_KEY)
+            .unwrap()
```
bulenttastan, line 29: We should use `.expect` with proper error messages.

rocksdb/src/moved.rs
```diff
@@ -0,0 +1,108 @@
...
+    fn tree(&self) -> EthTrie<RocksEthTrieDb<'db>> {
+        let db = self.db.clone();
+
+        match self.state_root {
+            B256::ZERO => EthTrie::new(db),
+            root => EthTrie::from(db, root).unwrap(),
+        }
+    }
+
+    fn insert_change_set_into_merkle_trie(
```
bulenttastan, line 53: Maybe we should call this fn `persist_state_change` to keep it consistent with `persist_state_root`. Plus the insertion into Merkle trie is implied.

rocksdb/src/move.rs
```diff
@@ -0,0 +1,100 @@
...
+        &self,
+        address: &AccountAddress,
+        _struct_tag: &StructTag,
+        _metadata: &[Metadata],
+        _layout: Option<&MoveTypeLayout>,
+    ) -> Result<(Option<Bytes>, usize), Self::Error> {
+        let key = address.to_canonical_string();
+        let db = self.db();
```
bulenttastan, line 62: nit: We don't need this line. We can just `self.db.get...` on the next line.

rocksdb/src/move.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+impl<'db> ResourceResolver for RocksEthTrieDb<'db> {
+    type Error = rocksdb::Error;
+
+    fn get_resource_bytes_with_metadata_and_layout(
+        &self,
+        address: &AccountAddress,
+        _struct_tag: &StructTag,
```
bulenttastan, line 57: Just curious, why are these params needed but not used?

Cargo.toml
```diff
@@ -50,11 +51,13 @@ move-vm-test-utils = { git = "https://github.com/aptos-labs/aptos-core", tag = "
 move-vm-types = { git = "https://github.com/aptos-labs/aptos-core", tag = "aptos-node-v1.14.0" }
 moved = { path = "moved" }
 moved-engine-api = { path = "engine-api" }
+moved-rocksdb = { path = "rocksdb" }
 once_cell = "1.19"
 op-alloy = { version = "0.6", features = ["full", "std", "k256", "serde"] }
 openssl = "0.10"
 regex = "1.11.1"
 revm = { version = "17", features = ["optional_balance_check"] }
+rocksdb = { git = "https://github.com/rust-rocksdb/rust-rocksdb", rev = "7f9cba4" }
```
birchmd, line 60: Why use a GitHub dependency instead of a regular crates.io one? The crate is [listed here](https://crates.io/crates/rocksdb).

moved/src/storage.rs
```diff
@@ -29,23 +29,26 @@ pub trait State {
...
-    /// using the [`move_table_extension`].
+    /// Applies the `changes` to the blockchain state. In addition, applies `table_changes`
+    /// from the [`move_table_extension`].
     fn apply_with_tables(
         &mut self,
         changes: ChangeSet,
         table_changes: TableChangeSet,
     ) -> Result<(), Self::Err>;
 
+    /// Returns a reference to a [`TreeReader`] that can access the merkle trie holding the current
```
birchmd, line 43:
```suggestion
    /// Returns a reference to a [`DB`] that can access the merkle trie holding the current
```

rocksdb/src/lib.rs
birchmd: I agree a different name is needed here. I don't like using reserved words even if we can do it via the `r#` trick.

rocksdb/src/move.rs
birchmd: The name comes from the `ResourceResolver` trait, we cannot change it.

rocksdb/src/move.rs
```diff
@@ -0,0 +1,100 @@
...
+            if let Ok(module) = CompiledModule::deserialize(&bytes) {
+                return module.metadata;
+            }
+        }
+
+        Vec::new()
+    }
+
+    fn get_module(&self, id: &ModuleId) -> Result<Option<Bytes>, Self::Error> {
+        let key = id.address().to_canonical_string();
```
birchmd, line 44: As discussed on the call, we'll need to do something to distinguish different kinds of keys.

For this particular case the key you are using is not enough because a single address can have multiple different modules. This is why `ModuleId` also includes a `name` field.

But all of this is secondary to the point that I don't think the resolver traits are needed for the `RocksDB` type because we will only be accessing the storage via the trie. So only the `HistoricResolver` needs to implement the resolver traits.

rocksdb/src/move.rs
```diff
@@ -0,0 +1,100 @@
...
+    ) -> Result<Option<Bytes>, PartialVMError> {
+        todo!()
+    }
+}
+
+impl<'db> DB for RocksEthTrieDb<'db> {
+    type Error = rocksdb::Error;
+
+    fn get(&self, key: &[u8]) -> Result<Option<Vec<u8>>, Self::Error> {
+        self.db().get(key)
```
birchmd, line 86: Reminder about key discriminator.

rocksdb/src/moved.rs
birchmd: This function looks nearly identical to the one for the in-memory implementation. I wonder if we can write one implementation shared by both rocksdb and in-mem by having it take the `EthTrie` as a parameter instead of `self`.

birchmd (commented):
Nice start! I've left a few comments.

Cargo.toml
RomanHodulak: This one has a newer RocksDB 9.10, which is not in the published crates.

rocksdb/src/move.rs
RomanHodulak: It's part of the `DB` trait from `eth_trie` crate, making it necessary to implement. It's supposed to perform all pending writes from in-memory cache into files.

It could be used internally or triggered from the public API.

birchmd (approved):
Nice, looks good to me!

PR #190, opened 2025-01-17
Feat(trie): Match Ethereum account structure for EVM resources

birchmd:
### Description
More changes to how the trie works for EVM-related resources. The code now 100% reproduces what Ethereum does and therefore should always produce merkle proofs that is accepted by Ethereum tooling.

### Changes
- EVM account data stored in the trie in RLP-encoded format
- Each EVM account had its own storage trie, these are still included in our main state trie though.

### Testing
- Existing tests.
- Extra checks in `eth_getProof` integration test.

### Notes
- I still do not handle self-destruct for EVM changes. Previously there was not a good way to do it since EVM account storage was spread out over multiple resources, but now that each account has a single resource the implementation should be straightforward. I left it out of scope for this PR, but it can be done in a future PR. It's not pressing because none of the L2 contracts use self-destruct.


moved/src/move_execution/evm_native/trie_types.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub fn serialize(&self) -> Vec<u8> {
+        let db_bytes = self.trie.db.serialize();
+        [self.root_hash.as_slice(), &db_bytes].concat()
+    }
+
+    pub fn try_deserialize(bytes: &[u8]) -> Option<Self> {
+        if bytes.len() < 32 {
+            return None;
+        }
+        let root_hash = B256::from_slice(&bytes[..32]);
```
bulenttastan, line 73: nit: I feel like these next few lines needs a comment explaining what's happening.

bulenttastan (commented):
It looks good. Leaving @RomanHodulak to give another look.

I tested this PR with the integration test and the Proposer is happy with the proof returned.

moved/src/move_execution/evm_native/native_evm_context.rs
```diff
@@ -93,21 +122,18 @@ impl<'a> DatabaseRef for ResolverBackedDB<'a> {
...
-                value.value_as::<u256::U256>()?.to_u256()
-            }
-            None => {
-                // Zero is the default value when there is no entry
-                return Ok(U256::ZERO);
-            }
-        };
+        let mut cache_lock = self
+            .storage_cache
+            .write()
```
RomanHodulak, line 127: If we only use write lock, then we don't benefit from using an RwLock.

Although I don't think we are going to have multiple readers in any case.

RomanHodulak (approved):
Incredible that you made it match.

moved/src/types/state.rs
```diff
@@ -392,6 +393,28 @@ impl KeyHashable for TreeKey {
     }
 }
 
+/// Type representing the keys used in the state trie.
```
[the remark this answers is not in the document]
birchmd: Yes, good catch. Fixed in https://github.com/MovedNetwork/op-move/pull/190/commits/2e778d0043f7fb235e9deb88d73b7070a8afe675

moved/src/move_execution/evm_native/native_evm_context.rs
birchmd: As discussed on the call, the `RwLock` is primarily giving us interior mutability for the cache as opposed to multi-threaded access. I have a comment describing the purposes of the the cache and lock in https://github.com/MovedNetwork/op-move/pull/190/commits/2e778d0043f7fb235e9deb88d73b7070a8afe675

moved/src/move_execution/evm_native/trie_types.rs
birchmd: Comment added in https://github.com/MovedNetwork/op-move/pull/190/commits/d43473994f8980fe4416b7a00557fe35e0fabc46

bulenttastan (approved):

PR #191, opened 2025-01-18
test(storage): Test that there is no collision between different column families

RomanHodulak:
### Description
Introduces integration tests into the `storage` crate.

### Changes
- Initialize directory for integration tests in `storage` trait
- Declare `common` module with code for creating empty database for testing
- Add test that interacts with file system using `rocksdb` that simulates the situation where the same key used for storing *state root* is used through the `DB` trait of the `eth_trie` crate and test that nothing was overwritten.

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### DIFF
--- a/storage/Cargo.toml
+++ b/storage/Cargo.toml
@@ -11,3 +11,6 @@ move-core-types.workspace = true
 move-table-extension.workspace = true
 moved.workspace = true
 rocksdb.workspace = true
+
+[dev-dependencies]
+hex-literal.workspace = true
--- a/storage/tests/collision.rs
+++ b/storage/tests/collision.rs
@@ -1,0 +1,40 @@
+use {
+    eth_trie::DB,
+    hex_literal::hex,
+    moved::primitives::B256,
+    moved_storage::{RocksEthTrieDb, ROOT_KEY},
+};
+
+mod common;
+
+#[test]
+fn test_column_families_do_not_collide() {
+    let rocks = common::create_db();
+    let db = RocksEthTrieDb::new(&rocks);
+
+    let random_32_bytes = B256::new(hex!(
+        "50596cee391a497683672d9396379f56cd8e96476b844557933f48039c483a81"
+    ));
+
+    db.put_root(random_32_bytes).unwrap();
+
+    // We assume that `put_root` uses `ROOT_KEY` to store the state root under the hood
+    // If the keys are not separated by column families, then this overwrites the state root
+    let random_value = hex!("feef");
+
+    db.insert(ROOT_KEY.as_bytes(), random_value.as_slice().to_vec())
+        .unwrap();
+
+    let expected_root = random_32_bytes;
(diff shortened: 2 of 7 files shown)


storage/Cargo.toml
```diff
@@ -11,3 +11,6 @@ move-core-types.workspace = true
 move-table-extension.workspace = true
 moved.workspace = true
 rocksdb.workspace = true
+
+[dev-dependencies]
+hex-literal = "0.4"
```
bulenttastan, line 16: nit: probably should move to workspace

storage/tests/collision.rs
```diff
@@ -0,0 +1,32 @@
+mod common;
```
bulenttastan, line 1: nit: We usually keep `use` before `mod`.

storage/tests/collision.rs
```diff
@@ -0,0 +1,32 @@
...
+    moved_storage::{RocksEthTrieDb, ROOT_KEY},
+};
+
+#[test]
+fn test_column_families_do_not_collide() {
+    let rocks = common::create_db();
+    let db = RocksEthTrieDb::new(&rocks);
+
+    let expected_root = B256::new(hex!(
+        "50596cee391a497683672d9396379f56cd8e96476b844557933f48039c483a81"
```
bulenttastan, line 16: Is this the state root after adding `0xfeef`? If so can you leave a short comment above describing what this root value is.

storage/tests/collision.rs
```diff
@@ -0,0 +1,32 @@
...
+    ));
+    db.put_root(expected_root).unwrap();
+
+    let expected_value = hex!("feef");
+    db.insert(ROOT_KEY.as_bytes(), expected_value.as_slice().to_vec())
```
bulenttastan, line 21: I think we should leave  a comment saying, using the same key (`ROOT_KEY`) for root and tree element should not cause any collision.

storage/Cargo.toml
ilchu: As far as I can tell, we were already using `alloy::hex`, at least in tests. Seems to be the same functionality

storage/tests/collision.rs
ilchu: I feel it would be a pain to maintain this, as with every change of the state root this test will need to be updated too. Maybe it warrants a separate issue for making it more ergonomic?

RomanHodulak: It's random bytes, trie logic does not affect this test.

storage/Cargo.toml
RomanHodulak: It's the same library that `alloy` uses (`alloy::hex` is a re-export from `hex-literal`), so this will not add any additional work to the compiler, and we don't need to bring in entire `alloy`.

bulenttastan (approved):

PR #192, opened 2025-01-20
feat(storage): Implement block repository

RomanHodulak:
### Description
Implements storing blocks inside `rocksdb`.

### Changes
- Added `BlockRepository` implementation backed by `rocksdb`
- Added new column family for blocks into `rocksdb` so that they have separate isolated namespace to avoid collision with keys for other purposes
- Added test for column family name uniqueness

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6907,6 +6907,7 @@ dependencies = [
 name = "moved-storage"
 version = "0.1.0"
 dependencies = [
+ "bcs 0.1.3",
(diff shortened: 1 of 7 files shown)


birchmd (approved):
Nice, looks good!

I think we should also have a column to store the block height to block hash mapping.

PR #193, opened 2025-01-21
refactor: Resolve `clippy::too_many_args` warning for transaction execution

RomanHodulak:
### Description
Changes the design of our transaction execution entry points by making them use only a single argument.

### Changes
- Introduce the concept of an input struct for executing transactions
- Use the input structs for deposit and for canonical transaction execution
- Update all usages
- Clear `#[allow(clippy::too_many_args)]` on the affected places

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### DIFF
--- a/moved/src/move_execution/mod.rs
+++ b/moved/src/move_execution/mod.rs
@@ -16,7 +16,7 @@ use {
         primitives::{ToEthAddress, B256},
         types::{
             session_id::SessionId,
-            transactions::{NormalizedExtendedTxEnvelope, TransactionExecutionOutcome},
+            transactions::{DepositedTx, NormalizedEthTransaction, TransactionExecutionOutcome},
         },
     },
     alloy::primitives::{Bloom, Keccak256, Log, LogData},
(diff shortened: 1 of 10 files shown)


bulenttastan:
I was hoping to create segmented input params like `VMInputs`, `TransactionInputs` etc, but it seems almost impossible to segment the inputs in a reusable way for both execution and verification. So I think this seems like a reasonable way to keep all the inputs in a single struct.

moved/src/move_execution/mod.rs
```diff
@@ -116,33 +116,76 @@ where
...
-    state: &(impl MoveResolver<PartialVMError> + TableResolver),
-    genesis_config: &GenesisConfig,
-    l1_cost: u64,
-    l2_fee: impl L2GasFee,
-    l2_input: L2GasFeeInput,
-    base_token: &impl BaseTokenAccounts,
-    block_header: HeaderForExecution,
+pub enum TransactionExecutionInput<
+    'input,
+    S: MoveResolver<PartialVMError> + TableResolver,
```
birchmd, line 121: Typically you don't put the trait bounds on the data type definition only on the function/method implementations. The reason is because trait bounds restrict _behaviour_, which is the domain of methods/functions, where as data type definitions are about _structure_ only. Keeping trait bounds only on behaviour definitions allows the opportunity to reuse data types where the same structure makes sense in multiple different behaviour settings.

birchmd (approved):
Thanks for addressing this! One comment to look at before merging please.

RomanHodulak:
Thanks for taking a look. I think a flat structure in this situation makes sense.

PR #195, opened 2025-01-21
Fix: make sure deposits work

birchmd:
### Description
For op-node to recognize the deposits as going through we must include a non-null withdrawals field. A few other bugs I noticed while debugging this are also addresses. With this change the deposits seem to be working flawlessly!

### Changes
- Non-null withdrawals
- Non-null deposit version and nonce
- Drop duplicate transactions
- Not all deposit-type transactions are actual deposits

### Testing
- Integration test

### Notes
- This raises the priority to audit the other TODOs related to `None` fields. They may be causing other problems we don't know about.


moved/src/types/state.rs
```diff
@@ -484,3 +532,16 @@ impl WithPayloadAttributes for Header {
         }
     }
 }
+
+#[test]
+fn test_get_deposit_nonce() {
+    const INPUT: [u8; 420] = alloy::hex!("d764ad0b0001000000000000000000000000000000000000000000000000000000000002000000000000000000000000c8088d0362bb4ac757ca77e211c30503d39cef4800000000000000000000000042000000000000000000000000000000000000100000000000000000000000000000000000000000000000056bc75e2d631000000000000000000000000000000000000000000000000000000000000000030d4000000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000000a41635f5fd000000000000000000000000c152ff76a513e15be1be43d102a881f076e707b3000000000000000000000000c152ff76a513e15be1be43d102a881f076e707b30000000000000000000000000000000000000000000000056bc75e2d631000000000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000");
```
RomanHodulak, line 538: :see_no_evil:

RomanHodulak (approved):
Great stuff, those optional fields are really not that optional.

bulenttastan (approved):
👏

PR #196, opened 2025-01-21
perf: Build genesis state at compile time reducing tests time by 90%

RomanHodulak:
### Description
Creates an image of the genesis state that gets restored every time the tests are run.

This results in a decrease of the time it takes to run all tests by about 90-95% from over 3 minutes to about 10 seconds (excluding the build time + the integration `test_on_ethereum` test in both cases).

The image is dumped into a file, which is then embedded in a dedicated crate binary at compile time. In this way the runtime does not even have to do a file read, instead it directly reads from a variable.

Not only does the genesis state generation moves from runtime to compile time but it also happens only once per build of the `moved-genesis-image` crate as opposed to once per test on every run.

### Changes

- Store genesis state in an image dumped into a file
- Restore the state from the image instead of generating it every time

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt


moved/src/genesis/mod.rs
```diff
@@ -5,13 +5,21 @@ use {
...
+mod cache;
 pub mod config;
 mod framework;
 mod l2_contracts;
 
 pub fn init(
     config: &GenesisConfig,
     state: &impl State<Err = PartialVMError>,
+) -> (ChangeSet, TableChangeSet) {
+    cache::try_load().unwrap_or_else(|| cache::save(config, state))
```
bulenttastan, line 17: Should we feature block cached data to tests only?

birchmd: I agree this feels like a tests-only implementation.

birchmd:
> what's left to improve is the way that the image is created, stored and restored.

And updated. We talked about the idea of using a build script to make sure the genesis cache file is invalidated when there are code changes that should cause it to be re-created. We'll have to be careful with this because we want to make sure it is always re-created when it needs to be, but obviously not when it doesn't need to be because otherwise that defeats the purpose of having a cache.

Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [workspace]
 resolver = "2"
 members = [
-    "engine-api",
-    "genesis-builder",
-    "moved",
-    "server",
+    "engine-api", "evm-ext", "genesis",
```
bulenttastan, line 4: Wow, this PR is a massive upgrade ✨

Why not each member go on to a new line?

Also curious, how come the unit test time on github action didn't improve? I'd think it would be improved too, but it still took 18 mins.

RomanHodulak: Thanks for reviewing!

I added the crates via a command so I didn't notice.

From the latest job runs I can tell that it took 4-5 minutes off the job's time. However, the vast majority of it is not spent on running the tests but rather compiling + downloading dependencies, that's why the difference gets lost in that great amount of time.

bulenttastan: Ok just noticed the 22 min vs 18 min difference. We're making this change for local testing instead of Github improvement. So this work is appropriate 🙌

primitives/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "moved-primitives"
```
bulenttastan, line 2: I'm not sure if primitives should be its own crate, but then again why not. @birchmd what do you think?

server/Cargo.toml
```diff
@@ -22,10 +22,15 @@ serde_json.workspace = true
 tokio.workspace = true
 warp-reverse-proxy.workspace = true
 warp.workspace = true
+moved-state.workspace = true
```
bulenttastan, line 25: nit: Alphabetical ordering. I noticed in other `Cargo.toml` files too.

genesis-image/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "moved-genesis-image"
+description = "Builds and embeds built image of genesis blockchain state into the binary for instanenious loading"
```
bulenttastan, line 3: nit: Typo  `instanenious` → `instantaneous`

genesis/src/genesis/framework.rs
```diff
@@ -13,21 +12,23 @@ use {
...
     move_vm_types::gas::UnmeteredGasMeter,
+    moved_state::State,
     once_cell::sync::Lazy,
     std::{collections::BTreeMap, fs, path::PathBuf},
     sui_framework::SystemPackage,
     sui_types::base_types::ObjectID,
 };
 
-pub const CRATE_ROOT: &str = env!("CARGO_MANIFEST_DIR");
+pub const CRATE_ROOT: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/../moved");
```
bulenttastan, line 27: I saw this to get the workspace directory. It might be useful, but not necessary to use it here.
https://github.com/rust-lang/cargo/issues/3946#issuecomment-973132993

moved/src/move_execution/tests/utils.rs
```diff
@@ -1,4 +1,7 @@
-use {super::*, crate::types::transactions::NormalizedExtendedTxEnvelope};
+use {
+    super::*, crate::types::transactions::NormalizedExtendedTxEnvelope,
```
bulenttastan, line 2: I think `NormalizedExtendedTxEnvelope` got snuck in by accident before.
This should just be `use super::*;` and everything else would go into `mod.rs`.

server/src/lib.rs
```diff
@@ -138,9 +138,18 @@ pub fn initialize_state_actor(
     repository.add(&mut block_memory, genesis_block);
 
     let mut state = InMemoryState::new();
-    let (genesis_changes, table_changes) = genesis::init(&genesis_config, &state);
+    let (genesis_changes, table_changes) = {
```
bulenttastan, line 141: Is this more readable instead?
```rust
#[cfg(test)]
let (genesis_changes, table_changes) = moved_genesis_image::load();
#[cfg(not(test))]
let (genesis_changes, table_changes) = moved_genesis::build(&moved_genesis::MovedVm, &genesis_config, &state);
    
```

primitives/Cargo.toml
birchmd: I think this is done to prevent dependency cycles among our crates now that the big `moved` crate has been separated out (which was necessary to make the genesis image crate effective). So it makes sense to have primitives as separate crate now.

birchmd (approved):
Awesome job! Factoring the genesis image out into a separate crate with its own build file is a great idea!

genesis/src/genesis/framework.rs
RomanHodulak: Nice, well I prefer to keep it this way in this case because it works out of the box with no workarounds.

server/src/lib.rs
RomanHodulak: I think it's more readable to have only one place where the variable is declared as it is currently.

PR #197, opened 2025-01-22
feat(storage): Implement block queries backed by `rocksdb` as its persistent storage engine

RomanHodulak:
### Description
Extends the `rocksdb` integration with `moved` by block queries.

### Changes
- Implement `BlockQueries` for rocksdb backed storage engine
- Add new column family for storing height index
- Update height to hash index when adding blocks

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt


storage/src/block.rs
```diff
@@ -1,35 +1,92 @@
...
+    fn by_hash(
+        &self,
+        db: &Self::Storage,
+        hash: B256,
+        include_transactions: bool,
+    ) -> Option<BlockResponse> {
+        let cf = block_cf(db);
+        if include_transactions {
+            db.get_cf(&cf, hash)
+                .unwrap()
```
bulenttastan, line 50: Maybe we should use `.expect("")`s for all the .`unwraps()` here and below.

storage/src/block.rs
```diff
@@ -1,35 +1,92 @@
...
 
-    fn add(&mut self, storage: &mut Self::Storage, block: ExtendedBlock) {
-        let cf = Self::cf(storage);
+    fn add(&mut self, db: &mut Self::Storage, block: ExtendedBlock) {
+        let cf = block_cf(db);
         let bytes = bcs::to_bytes(&block).unwrap();
-        storage.put_cf(&cf, block.hash, bytes).unwrap()
+        db.put_cf(&cf, block.hash, bytes).unwrap();
+        let cf = height_cf(db);
+        db.put_cf(&cf, block.block.header.number.to_le_bytes(), block.hash)
```
bulenttastan, line 24: Does it matter if this is `to_le_bytes` or `to_be_bytes`? Big endian is more human readable.

moved/src/block/in_memory.rs
```diff
@@ -92,36 +92,37 @@ impl BlockRepository for InMemoryBlockRepository {
 pub struct InMemoryBlockQueries;
 
 impl BlockQueries for InMemoryBlockQueries {
+    type Err = ();
```
birchmd, line 95: I suggest using [Infallible](https://doc.rust-lang.org/std/convert/enum.Infallible.html) when there is no possibility for error.

storage/src/block.rs
```diff
@@ -1,35 +1,84 @@
...
+        db: &Self::Storage,
+        hash: B256,
+        include_transactions: bool,
+    ) -> Result<Option<BlockResponse>, Self::Err> {
+        let cf = block_cf(db);
+
+        Ok(db
+            .get_cf(&cf, hash)?
+            .and_then(|v| bcs::from_bytes(v.as_slice()).ok())
+            .map(if include_transactions {
```
birchmd, line 54: I this is slick! I'm surprised the type checker is able to handle it correctly.

birchmd (approved):

PR #199, opened 2025-01-23
Feat(integration-test): Test withdrawal flow

birchmd:
### Description
This PR does not change any main logic, only the integration tests are modified. This is a good thing because it means our logic for merkle proofs is already correct!

Some of the main changes to the integration test setup are with `op-proposer`. It turned out that we were not pushing much data to the L1 in the short time we run the test because `op-proposer` waits for blocks to be finalized by default (a sensible safety precaution) and only pushes updates every 120 L2 blocks (a sensible way to save money on L1 gas fees). In our test setting we know there will never be any fork, so we can push unfinalized blocks and we don't care about L1 gas so we can push as often as we like. Therefore I added `--allow-non-finalized` to the command which starts the proposer and changed our L2 config file to have `l2OutputOracleSubmissionInterval = 30`. Additionally, I included `--num-confirmations=1` in the proposer startup because we know the L1 isn't going to fork either in our test condition.

Another issue the proposer was having is that it expects the L1 to produce blocks regularly, but we are intentionally running geth in a mode where it only produces blocks when there are pending transactions (this is done so that deploying the rollup contracts on the L1 does not take ages). So I have added a background task that submits a transaction to the L1 every 3 seconds, ensuring that new blocks are regularly created.

Separately from the proposer, our rollup config was set up to use the (relatively new) fault proofs part of the OP-stack. This significantly complicates withdrawals (the benefit of this complexity is that users do not need to rely on a privileged proposer to perform the withdrawals; it can be done permissionlessly) and frankly I did not want to learn about the `DisputeGame` API when the logic for driving the withdrawals flow is already around 200 lines of code. So instead I set `useFaultProofs=false` in our config to use the old flow. In the future we will want to figure out how to run a node with fault proofs enabled (something to do with `op-challenger`...).

As eluded to above, the bulk of the new code added in this PR is for doing the withdrawal flow. This requires interacting with two of the rollup contracts on the L1 so I also added their ABI JSON files to our test resources in order to take advantage of alloy's `sol` macro. Hopefully the rest of the withdrawal flow code is commented well enough to be self-explanatory.

### Changes
- L2 config changes
- Proposer config changes
- New background task in integration test for submitting L1 transactions at regular intervals
- - New logic in integration test for withdrawal flow

### Testing
- It's all tests XD


server/src/tests/integration/withdrawal.rs
```diff
@@ -0,0 +1,212 @@
...
+//! (Separated into its own module because there is a lot of it.)
+
+use super::*;
+
+#[allow(clippy::too_many_arguments)]
+mod op_oracle {
+    alloy::sol!(
+        #[sol(rpc)]
+        L2OutputOracle,
+        "src/tests/res/L2OutputOracle.json"
```
bulenttastan, line 11: These contract ABI files are under `server/src/tests/optimism/packages/contracts-bedrock/snapshots/abi`.
Should we copy into `res` or refer to them?

server/src/tests/integration.rs
```diff
@@ -81,6 +91,15 @@ async fn test_on_ethereum() -> Result<()> {
     deploy_move_counter().await?;
 
     // 12. Cleanup generated files and folders
+    hb.abort();
```
bulenttastan, line 94: Maybe we can move the heartbeat cleanup logic into `heartbeat.rs` and call it with a single line here.

server/src/tests/integration/withdrawal.rs
birchmd: I think the file I have included in `res` has more information than what is in `snapshots/abi`. For example the `WithdrawalTransaction` struct definition.

server/src/tests/integration/withdrawal.rs
```diff
@@ -0,0 +1,212 @@
...
+            .saturating_to();
+
+        // Timeout to prevent this from being an infinite loop if something breaks
+        if now.elapsed().as_secs() > 10 * 60 {
+            anyhow::bail!(
+                "WITHDRAW ERROR: L1 contract `L2OutputOracleProxy` not updated to block containing withdraw within 10 minutes. CurrentBlock={l2_block_number} TargetBlock={withdraw_block_number}"
+            );
+        }
+
+        if l2_block_number < withdraw_block_number {
```
bulenttastan, line 88: nit:
```rust
if l2_block_number >= withdraw_block_number {
    break;
}
tokio::time::sleep(Duration::from_secs(OP_BRIDGE_IN_SECS)).await;
```

bulenttastan (approved):
I see it worked on the server, the account was deposited 100 ETH and ended with 99 ETH, since 1 ETH was withdrawn back to L1.

server/src/tests/integration/withdrawal.rs
```diff
@@ -0,0 +1,212 @@
...
+        let l2_block_number: u64 = l2_oracle_contract
+            .latestBlockNumber()
+            .call()
+            .await
+            .unwrap()
+            ._0
+            .saturating_to();
+
+        // Timeout to prevent this from being an infinite loop if something breaks
+        if now.elapsed().as_secs() > 10 * 60 {
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/199/commits/58e44ffd7d56df23c0c9f5a0fec846dadf6d2bf2

server/src/tests/integration/withdrawal.rs
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/199/commits/58e44ffd7d56df23c0c9f5a0fec846dadf6d2bf2

PR #209, opened 2025-01-26
refactor(api): Generalize JSON-RPC params parsing

RomanHodulak:
### Description
The endpoint implementation contains parameter parsing logic, that contains a large portion of duplicated code.

In this PR we use generic return types to generalize and remove these duplicated parsing functions.

Targets only cases where the argument count is exact with no further logic.

### Changes
- Remove duplicate custom parsing logic and generalize it into a generic version

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### DIFF
--- a/api/src/json_utils.rs
+++ b/api/src/json_utils.rs
@@ -34,3 +34,47 @@ pub fn access_state_error<E: fmt::Debug>(e: E) -> JsonRpcError {
 pub fn transaction_error<E: fmt::Debug>(e: E) -> JsonRpcError {
     JsonRpcError::without_data(3, format!("Execution reverted: {e:?}"))
 }
+
+pub fn parse_params_0(request: serde_json::Value) -> Result<(), JsonRpcError> {
+    let params = get_params_list(&request);
+    match params {
+        [] => Ok(()),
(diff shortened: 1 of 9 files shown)


api/src/json_utils.rs
```diff
@@ -34,3 +34,47 @@ pub fn access_state_error<E: fmt::Debug>(e: E) -> JsonRpcError {
 pub fn transaction_error<E: fmt::Debug>(e: E) -> JsonRpcError {
     JsonRpcError::without_data(3, format!("Execution reverted: {e:?}"))
 }
+
+pub fn parse_params_0(request: serde_json::Value) -> Result<(), JsonRpcError> {
```
ilchu, line 38: I feel it might call for a little `macro_rules!` here. Though AFAIK @bulenttastan is usually against it, but shouldn't hurt readability / maintainability too much. Also otherwise I would advocate for `parse_zero...`, `parse_one...` etc naming, but that's just a nit.

ilchu (commented):
Good time to abstract these things away, like the conciseness!

PR #210, opened 2025-01-26
feat(api): Add `eth_getTransactionByHash` endpoint

RomanHodulak:
### Description
Adds new JSON-RPC API endpoint for fetching transactions by hash.

Currently returning empty responses.

### Changes
- Introduce `eth_getTransactionByHash` to API
- Respond with empty messages on this endpoint

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt


moved/src/types/state.rs
```diff
@@ -228,6 +232,7 @@ impl From<Query> for StateMessage {
 }
 
 pub type RpcBlock = alloy::rpc::types::Block<op_alloy::rpc_types::Transaction>;
+pub type RpcTx = op_alloy::rpc_types::Transaction;
```
bulenttastan, line 235: Maybe we should call it `RpcTransaction`

moved/src/types/state.rs
```diff
@@ -370,6 +375,9 @@ fn compute_from(tx: &OpTxEnvelope) -> Result<Address, alloy::primitives::Signatu
     }
 }
 
+#[derive(Debug)]
+pub struct TransactionResponse(pub op_alloy::rpc_types::Transaction);
```
bulenttastan, line 379: Why isn't it similar to TransactionReceipt, like:
```rust
pub type TransactionResponse = op_alloy::rpc_types::Transaction;
```

birchmd (approved):
+1 for consistency with newtype vs type alias, but looks good otherwise.

moved/src/types/state.rs
RomanHodulak: I will keep the inconsistency with receipt in mind and look into this again as I will work throughout transactions and receipts.

---

There is added benefit of it being a local type.

Local types can have foreign traits implemented for them, such as `From` trait from foreign type.

The local type does not automatically inherit traits for its inner type that it wraps, instead it supports only the behavior that we intentionally mean for it to have.

Any overhead gets optimized away thanks to zero-cost abstractions.

PR #211, opened 2025-01-27
fix: Add compiler flag to test only imports

RomanHodulak:
### Description
Resolves warnings produced when compiling non-test builds

### Changes
- Add compiler flag to test only imports

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt


bulenttastan (approved):
Good catch.
I also think having the tests run faster helps us iterate faster.

PR #212, opened 2025-01-27
build(server): Rename target binary `moved_server` => `op-move`

RomanHodulak:
### Description
In case this is wanted, this PR changes the name of the output binary to `op-move` which should align nicely with the rest of the op stack.

When building for default target with release profile, the produced binary executable will be `./target/release/op-move`

### Changes
- Rename target binary `moved_server` => `op-move`

### Testing
```
cargo run --bin op-move
```


bulenttastan (approved):
Good change. This makes the main binary name match the project name.

PR #213, opened 2025-01-27
feat: Add transaction reading and writing

RomanHodulak:
### Description
* Adds append-only repository and query for fetching by hash.
* Allows for querying transactions using API.
* Includes in-memory implementation.

### Changes
- Add write and read model for transactions
- Add in-memory implementations
- Include the model in `StateActor`
- Return proper responses in `api` endpoint
- Fill-in all missing fields

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### Notes
In this version, the transactions are stored twice. Once in their own memory space and once as a part of blocks.

This can be mitigated, but doing so would increase the amount of changes in this high priority PR as well as time to finish and review it.

Therefore we can conclude that it's better to target it as a separate issue.

### DIFF
--- a/moved/src/transaction/in_memory.rs
+++ b/moved/src/transaction/in_memory.rs
@@ -1,0 +1,65 @@
+use {
+    crate::{
+        transaction::{Transaction, TransactionQueries, TransactionRepository},
+        types::state::TransactionResponse,
+    },
+    alloy::eips::eip2718::Encodable2718,
+    moved_shared::primitives::B256,
+    std::convert::Infallible,
+};
+
+#[derive(Debug)]
+pub struct TransactionMemory {
+    txs: std::collections::HashMap<B256, Transaction>,
+}
+
+impl TransactionMemory {
+    pub fn add(&mut self, tx: Transaction) {
+        self.txs.insert(tx.0.trie_hash(), tx);
+    }
+
+    pub fn by_hash(&self, hash: B256) -> Option<Transaction> {
+        self.txs.get(&hash).cloned()
+    }
+}
+
+#[derive(Debug)]
+pub struct InMemoryTransactionRepository;
+
+impl Default for InMemoryTransactionRepository {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl InMemoryTransactionRepository {
+    pub fn new() -> Self {
+        Self
+    }
+}
+
+#[derive(Debug)]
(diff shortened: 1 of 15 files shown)


moved/src/transaction/in_memory.rs
```diff
@@ -0,0 +1,53 @@
...
+        types::state::TransactionResponse,
+    },
+    alloy::eips::eip2718::Encodable2718,
+    moved_shared::primitives::B256,
+    std::convert::Infallible,
+};
+
+#[derive(Debug)]
+pub struct TransactionMemory {
+    txs: std::collections::HashMap<B256, Transaction>,
```
bulenttastan, line 13: We should call its full name: `transactions`

moved/src/types/state.rs
```diff
@@ -378,6 +380,76 @@ fn compute_from(tx: &OpTxEnvelope) -> Result<Address, alloy::primitives::Signatu
...
+    fn from(value: Transaction) -> Self {
+        let (deposit_nonce, deposit_receipt_version) = get_deposit_nonce(&value.0)
+            .map(|nonce| (Some(nonce.nonce), Some(nonce.version)))
+            .unwrap_or((None, None));
+
+        Self(op_alloy::rpc_types::Transaction {
+            inner: alloy::rpc::types::eth::Transaction {
+                inner: value.0,
+                // TODO: Solve missing fields
+                block_hash: None,
```
bulenttastan, line 441: Are these really not available? We should really figure these out, cause when we were debugging the deposit issue the `None` values caused so much headache.

moved/src/transaction/write.rs
```diff
@@ -0,0 +1,50 @@
+use {
+    alloy::eips::eip2718::Encodable2718, moved_shared::primitives::B256,
+    op_alloy::consensus::OpTxEnvelope, std::fmt::Debug,
+};
+
+#[derive(Debug, Clone)]
+pub struct ExtendedTransaction {
+    pub effective_gas_price: u128,
+    pub inner: OpTxEnvelope,
```
bulenttastan, line 9: nit: I'd have `inner` as the first field.

bulenttastan (approved):
Well done 👏

Like you described in the notes, this is nicely optimized to access the transactions quicker, and we can optimize the storage usage in the future.

PR #218, opened 2025-01-28
feat: Add `effective_gas_price` into transactions inside block responses

RomanHodulak:
### Description
Fills-in missing gas price in transactions that are a part of block responses.

The recently added transaction response contains the `effective_gas_price` so this PR unifies this code path.

It is a step closer to removing transactions from blocks altogether and only use their own dedicated storage.

### Changes
- Unify code paths for pure transaction responses and transaction responses inside blocks

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### DIFF
--- a/moved/src/transaction/in_memory.rs
+++ b/moved/src/transaction/in_memory.rs
@@ -26,6 +26,13 @@ impl TransactionMemory {
     pub fn by_hash(&self, hash: B256) -> Option<ExtendedTransaction> {
         self.transactions.get(&hash).cloned()
     }
+
+    pub fn by_hashes(&self, hashes: impl IntoIterator<Item = B256>) -> Vec<ExtendedTransaction> {
+        hashes
+            .into_iter()
+            .filter_map(|hash| self.by_hash(hash))
+            .collect()
+    }
 }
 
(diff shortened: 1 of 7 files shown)


moved/src/transaction/in_memory.rs
```diff
@@ -26,6 +26,12 @@ impl TransactionMemory {
     pub fn by_hash(&self, hash: B256) -> Option<ExtendedTransaction> {
         self.transactions.get(&hash).cloned()
     }
+
+    pub fn by_hashes(&self, hash: impl IntoIterator<Item = B256>) -> Vec<ExtendedTransaction> {
```
bulenttastan, line 30: nit: input param name should be plural: `hashes`

bulenttastan (approved):
Looks good with clear separation of concerns.
The PR title is a little misleading though with effective gas price.

RomanHodulak:
Thanks!

The title relates to this change: https://github.com/MovedNetwork/op-move/pull/218/files#diff-519d00e671e930057011fbf21424428aab399d6d078f5357ca756dc3944e8b78L291

PR #219, opened 2025-01-29
refactor: Separate model for receipts from state actor internals

RomanHodulak:
### Description
* Moves the receipts logic read and write model out of the state actor internals.
* Represents a ground work for future work such as storage integration.
* Creates in-memory implementation out of the pre-existing implementation.

### Changes
- Add read and write model for receipts
- Integrate with state actor
- Remove the implementation pieces from state actor

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### Notes
The implementation could use refactoring as well but is kept intact to not introduce too many changes at once.


bulenttastan (approved):
Solid work 🙌

PR #222, opened 2025-01-30
refactor: Change `TransactionWithReceipt` into `ExtendedReceipt`

RomanHodulak:
### Description
Cuts out dependency on block to construct receipt response. Also removes transaction from receipt and instead keeps only needed information.

The result of this is that the receipts are more cleanly isolated.

This is a ground work to make receipts serializable.

### Changes
* Rename `TransactionWithReceipt` to `ExtendedReceipt`
* Remove dependency on block
* Remove transaction

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### DIFF
--- a/moved/src/state_actor.rs
+++ b/moved/src/state_actor.rs
@@ -1,3 +1,4 @@
+use {alloy::primitives::TxKind, op_alloy::consensus::OpTxEnvelope};
 pub use {
     payload::{NewPayloadId, NewPayloadIdInput, StatePayloadId},
     queries::{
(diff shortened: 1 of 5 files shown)


moved/src/state_actor.rs
```diff
@@ -596,19 +599,33 @@ where
                 U256::from(outcome.gas_used).saturating_mul(normalized_tx.tip_per_gas(base_fee)),
             );
 
-            receipts.push(TransactionWithReceipt {
-                tx_hash,
-                tx: tx.into(),
-                normalized_tx,
+            let (to, from) = match &normalized_tx {
```
bulenttastan, line 602: nit: Just a thought, I wonder if we can add a utility method to `NormalizedExtendedTxEnvelope` to get `(to, from)` tuple. I've seen this `match` case in multiple places.

bulenttastan (approved):
Let's ship it. Getting closer to persistent storage day by day.

PR #262, opened 2025-01-30
Clean: Remove unnecessary TODOs in SDK

bulenttastan:
### Description
These TODOs are not necessary, because the current way of reading the files is quite good. It doesn't need a result based interface.

### Changes
- Remove unnecessary `TODO` items


RomanHodulak (approved):
Quite good :pinched_fingers:

PR #263, opened 2025-01-31
feat(server): Integrate dependencies that use `rocksdb` as a backing storage engine

### LINKED ISSUE #216: Persist transactions and receipts in Rocks DB storage
The suspicion is in-memory storage causing high CPU usage for state persistence. This should be resolved with persistent storage.

RomanHodulak:
Closes #216 

### Description
Includes all the newly created dependencies implemented using `rocksdb`.

Allows switching between in-memory and rocksdb configuration using `storage` feature flag

### Changes
- Integrate dependencies that use `rocksdb` as a backing storage engine
- Allow switching between in-memory and rocksdb configuration using `storage` feature flag

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### Notes

You may now launch the node
* Using in-memory:
```
cargo run --bin op-move
```
* Using rocksdb:
```
cargo run ---bin op-move --features storage
```

### DIFF
--- a/server/src/dependency.rs
+++ b/server/src/dependency.rs
@@ -8,6 +8,7 @@ use {
     },
     moved_genesis::config::GenesisConfig,
     moved_state::State,
+    std::sync::Arc,
 };
 
 #[cfg(feature = "storage")]
@@ -56,7 +57,14 @@ pub fn block_repository() -> impl BlockRepository<Storage = SharedStorage> + Sen
 }
 
 pub fn state() -> impl State<Err = PartialVMError> + Send + Sync + 'static {
-    moved_state::InMemoryState::new()
+    #[cfg(feature = "storage")]
+    {
+        moved_storage::RocksDbState::new(Arc::new(moved_storage::RocksEthTrieDb::new(db())))
+    }
+    #[cfg(not(feature = "storage"))]
+    {
+        moved_state::InMemoryState::new()
+    }
 }
 
 pub fn state_query(genesis_config: &GenesisConfig) -> StateQueries {
--- a/server/src/dependency.rs
+++ b/server/src/dependency.rs
@@ -8,7 +8,6 @@ use {
     },
     moved_genesis::config::GenesisConfig,
     moved_state::State,
-    std::sync::Arc,
 };
 
 #[cfg(feature = "storage")]
@@ -59,7 +58,9 @@ pub fn block_repository() -> impl BlockRepository<Storage = SharedStorage> + Sen
 pub fn state() -> impl State<Err = PartialVMError> + Send + Sync + 'static {
     #[cfg(feature = "storage")]
     {
-        moved_storage::RocksDbState::new(Arc::new(moved_storage::RocksEthTrieDb::new(db())))
+        moved_storage::RocksDbState::new(std::sync::Arc::new(moved_storage::RocksEthTrieDb::new(
+            db(),
+        )))
     }
     #[cfg(not(feature = "storage"))]
     {
--- a/server/src/dependency.rs
+++ b/server/src/dependency.rs
@@ -1,0 +1,169 @@
+use {
+    move_binary_format::errors::PartialVMError,
+    moved::{
+        block::{BlockHash, BlockQueries, BlockRepository, MovedBlockHash},
+        move_execution::{BaseTokenAccounts, MovedBaseTokenAccounts},
+        receipt::{ReceiptQueries, ReceiptRepository},
+        transaction::{TransactionQueries, TransactionRepository},
+    },
+    moved_genesis::config::GenesisConfig,
+    moved_state::State,
+};
+
+#[cfg(feature = "storage")]
+pub type SharedStorage = &'static moved_storage::RocksDb;
+#[cfg(not(feature = "storage"))]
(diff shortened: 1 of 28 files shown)


server/src/dependency.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+pub fn base_token(
+    genesis_config: &GenesisConfig,
+) -> impl BaseTokenAccounts + Send + Sync + 'static {
+    MovedBaseTokenAccounts::new(genesis_config.treasury)
+}
+
+pub fn memory() -> SharedStorage {
+    #[cfg(feature = "storage")]
```
bulenttastan, line 37: I think we should make the RocksDB storage the default and keep in-memory as an option with feature named `memory`. Going forward the DB will be the default choice anyways.

We should see how the tests are affected by this. If tests are misbehaving with the DB, then we can run tests with the `--features memory` feature.

RomanHodulak: That does make sense. However, from a performance standpoint, the `rocksdb` adds additional dependencies that affect the compilation time significantly.

Turning it on explicitly makes for faster compilation time by default. The `memory` flag would need to be used in conjuction with `--no-default-features` to get that same benefit.

We will probably add a configuration option for the same purpose too, that can alert you if the compiled binary has improper flags set.

storage/src/generic.rs
```diff
@@ -23,3 +31,15 @@ macro_rules! int_impl {
...
+
+impl<T: serde::Serialize> ToValue for T {
+    fn to_value(&self) -> Vec<u8> {
+        serde_json::to_vec(self).unwrap()
+    }
+}
+
+impl<'de, T: serde::Deserialize<'de>> FromValue<'de> for T {
+    fn from_value(slice: &'de [u8]) -> Self {
+        serde_json::from_slice(slice).unwrap()
```
bulenttastan, line 43: 🚨 Making a deposit request on integration test breaks the storage integrity. It throws out the following error of not being able to construct `MaybeTaggedTxEnvelope`:
```
thread 'tokio-runtime-worker' panicked at storage/src/generic.rs:43:39:
called `Result::unwrap()` on an `Err` value: Error("data did not match any variant of untagged enum MaybeTaggedTxEnvelope", line: 1, column: 565)
```

and the `op-node` starts giving error:
```
t=2025-01-30T23:49:09-0500 lvl=warn msg="Failed to get payload" payload_id=0x03b5844e3da9cab0 payload_id=0x03b5844e3da9cab0 err="Failed to access state: SendError { .. }"
t=2025-01-30T23:49:09-0500 lvl=error msg="sequencer failed to seal block with unclassified error" err="failed to complete building block: error (1): failed to complete building on top of L2 chain 0x6148bc476b37448b722a7fed9688cfbf7921efb02e3c247f1d3d4116371b3778:460, id: 0x03b5844e3da9cab0, error (1): failed to get execution payload: unrecognized rpc error: Failed to access state: SendError { .. }"
t=2025-01-30T23:49:09-0500 lvl=error msg="cancelling old block sealing job" payload=0x03b5844e3da9cab0
```

server/src/dependency.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+pub fn state() -> impl State<Err = PartialVMError> + Send + Sync + 'static {
+    moved_state::InMemoryState::new()
+}
+
+pub fn state_query(genesis_config: &GenesisConfig) -> StateQueries {
+    #[cfg(feature = "storage")]
+    {
+        moved::state_actor::InMemoryStateQueries::from_genesis(genesis_config.initial_state_root)
```
bulenttastan, line 65: Is this supposed to be `moved_sorage::state::RocksDbState`?

RomanHodulak: Currently there is no such implementation.

But it is not strictly required, it only stores a single 32-byte hash per block.

storage/src/generic.rs
RomanHodulak: The deserialization used the wrong type. I pushed a change that uses the proper type explicitly.

server/src/dependency.rs
bulenttastan: Maybe we should make this change too while we're at it. I noticed the CPU performance issue lingering still. We just want to completely rule out lack of storage causing it.

RomanHodulak: You're right the `moved_storage::state::RocksDbState` was not actually used as it should be.

I just added it.

bulenttastan (approved):

server/src/dependency.rs
```diff
@@ -0,0 +1,178 @@
...
+    #[cfg(not(feature = "storage"))]
+    {
+        moved_state::InMemoryState::new()
+    }
+}
+
+pub fn state_query(genesis_config: &GenesisConfig) -> StateQueries {
+    #[cfg(feature = "storage")]
+    {
+        moved::state_actor::InMemoryStateQueries::from_genesis(genesis_config.initial_state_root)
```
bulenttastan, line 74: I think we should save the state roots in the persistent storage too, so that when we make the `op-move` restartable it won't lose state root history.

PR #264, opened 2025-01-31
refactor(api): Remove V2 engine methods and schema

### LINKED ISSUE #48: Remove *V2 RPC methods
We don't need to support these after all.

Code reference: https://github.com/MovedNetwork/op-move/blob/79df431c0deb8cfe59303d13adff50cbff10ec62/moved/src/main.rs#L285

RomanHodulak:
Closes #48 

### Description
Removes V2 engine methods and schema.

These were never implemented and not needed to be.

### Changes
- Removes V2 engine method names
- Removes V2 engine schema structs

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### DIFF
--- a/api/src/method_name.rs
+++ b/api/src/method_name.rs
@@ -2,9 +2,6 @@ use {crate::jsonrpc::JsonRpcError, std::str::FromStr};
 
 #[derive(Debug)]
 pub enum MethodName {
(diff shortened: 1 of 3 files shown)


bulenttastan (approved):
Good call on cleaning them up.

PR #271, opened 2025-02-04
feat(storage): Implement `StateQueries` using `rocksdb`

RomanHodulak:
### Description
Adds new `StateQueries` implementation backed by `rocksdb`

### Changes
- Adds new `StateQueries` implementation backed by `rocksdb`
- Adds new `OnTxBatch` callback specific to the `rocksdb` implementation of `StateQueries`
- Moves some trait types into type aliases to avoid clippy warning about complex types
- Uses this implementation when building with `--features storage`.

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### DIFF
--- a/moved/src/state_actor.rs
+++ b/moved/src/state_actor.rs
@@ -20,7 +20,6 @@ use {
         receipt::{ExtendedReceipt, ReceiptQueries, ReceiptRepository},
         transaction::{ExtendedTransaction, TransactionQueries, TransactionRepository},
         types::{
-            queries::ProofResponse,
             state::{
                 Command, ExecutionOutcome, Payload, PayloadId, PayloadResponse, Query,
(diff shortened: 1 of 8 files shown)


moved/src/state_actor.rs
```diff
@@ -94,14 +94,14 @@ pub struct StateActor<S, P, H, R, G, L1G, L2G, B, Q, M, SQ, T, TQ, N, RR, RQ> {
     execution_payloads: HashMap<B256, PayloadResponse>,
     pending_payload: Option<(PayloadId, PayloadResponse)>,
     mem_pool: HashMap<B256, (ExtendedTxEnvelope, L1GasFeeInput)>,
-    state: S,
+    pub state: S,
```
bulenttastan, line 97: Should we add setters instead of making them public? Since others aren't public either

bulenttastan (approved):

PR #274, opened 2025-02-06
ERC20 support

### LINKED ISSUE #185: Design: ERC-20 deposits
Need to figure out how ERC-20 deposits work:

https://github.com/MovedNetwork/op-move/blob/main/moved/src/move_execution/deposited.rs#L90

Goal: OP stack ERC-20 deposits should work for Moved identically to how they work in optimism.

Question: how to represent ERC-20 contracts in MoveVM?

ilchu:
### Description

Move contract generation for ERC20 from Optimism's superchain token list. Will close #185.

### Changes

- [x] Auto-generated (very similar, very repetitive) contracts for each coin. As the only things that differ are literally asset name, asset symbol and other metadata, it would probably be better to shift to a generic version of an ERC-20 contract in one copy. Then only declaring a few constants and maybe concretized versions of generic functions in each respective token's module will be needed
- [x] Fixed genesis builder not being runnable from directories other than `op-move` root
- [ ] Make `move_execution/eth_token.rs` functionality generic (i.e. allow for other implementations besides the default `EthToken` as of now)
- [ ] Filter the appropriate deposit events, sieve through ethereum logs, generalize minting
- [ ] Integrate into integration test :)

### Testing

Auto-generated tests in each module don't serve any actual purpose yet and they will probably be eventually discarded. A more sane way of testing would be to include the ERC-20 flow into our big `test_on_ethereum` integration workflow.


bulenttastan:
We can either delete this PR or use it on one of the relevant issues: #309, #310

ilchu:
> @ilchu We can either delete this PR or use it on one of the relevant issues: #309, #310

Yeah, I believe it will be relevant for #309, but I'll close this one for now.

PR #276, opened 2025-02-10
build: Dockerize Optimism stack

RomanHodulak:
### Description
Packages the blockchain network environment as a stack of docker services.

### Changes
* Add Dockerfiles to build images
* Add `docker-compose.yml` to build and run containers

### Testing
To deposit on L1:
```
docker/host/deposit.sh
```
Outputs transaction hash

To check balance on L2:
```
docker/host/balance.sh
```
Outputs the balance in Wei

### Notes
To start up:
```
docker compose down
docker compose build
docker compose up -d
```


docker/Dockerfile.geth
```diff
@@ -0,0 +1,47 @@
...
+
+# Switch to clean container
+FROM alpine:3.21.2
+WORKDIR /volume
+
+# Copy built binary from build image
+COPY --from=build /volume/build/bin/geth /volume/geth
+
+# Set entrypoint to geth binary
+ENTRYPOINT ["/volume/geth",
```
bulenttastan, line 31: While testing this, I noticed each line in the entrypoint needs a terminating character `\`
```
ENTRYPOINT ["/volume/geth", \
    "--dev", \
```

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+  geth:
+    image: moved-geth:latest
+    build:
+      context: ./docker
+      dockerfile: ./Dockerfile.geth
+    networks:
+      - localnet
+
+networks:
+  localnet:
```
bulenttastan, line 57: Will we open up the `58138` and `8545` ports here to expose L1 and L2?

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+      - localnet
+
+  geth:
+    image: moved-geth:latest
+    build:
+      context: ./docker
+      dockerfile: ./Dockerfile.geth
+    networks:
+      - localnet
+
```
bulenttastan, line 57: We can open up the port for `geth` via:
```
    ports:
      - "58138:58138"
```

docker/Dockerfile.geth
```diff
@@ -0,0 +1,47 @@
...
+    "--http",
+    "--http.addr",
+    "0.0.0.0",
+    "--http.port",
+    "58138",
+    "--http.corsdomain",
+    "*",
+    "--http.api",
+    "web3,debug,eth,txpool,net,engine"
+]
```
bulenttastan, line 47: We should expose the port via:
```
EXPOSE 58138
```

ilchu: And maybe respective ports for the op stack too? (e.g. 8560 for the proposer and so on)

docker/Dockerfile.optimism
```diff
@@ -0,0 +1,57 @@
...
+
+ARG TARGETARCH
+WORKDIR /opt
+
+RUN apk add clang lld curl build-base linux-headers git \
+    && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > rustup.sh \
+    && chmod +x ./rustup.sh \
+    && ./rustup.sh -y
+
+RUN [[ "$TARGETARCH" = "arm64" ]] && echo "export CFLAGS=-mno-outline-atomics" >> $HOME/.profile || true
```
ilchu, line 11: So I suppose we're just trying to replicate Bulent's setup as of now, or any other reason for ARM I'm missing?

docker/Dockerfile.geth
```diff
@@ -0,0 +1,47 @@
+FROM alpine:3.21.2 AS build
```
ilchu, line 1: Hope Alpine's musl doesn't introduce any linking problems :thinking:

docker-compose.yml
```diff
@@ -0,0 +1,78 @@
...
+      context: ./
+      dockerfile: ./docker/Dockerfile.op-geth
+    networks:
+      - localnet
+    depends_on:
+      - optimism
+    volumes:
+      - ./docker/volumes/contracts-bedrock:/volume/packages/contracts-bedrock
+
+  op-move:
```
bulenttastan, line 58: `op-move` depends on `op-geth`

docker/Dockerfile.optimism
```diff
@@ -0,0 +1,57 @@
...
+
+WORKDIR /opt/foundry
+RUN git clone https://github.com/foundry-rs/foundry.git . && git checkout 1bcd17c
+
+# see <https://github.com/foundry-rs/foundry/issues/7925>
+RUN git update-index --force-write-index
+
+RUN --mount=type=cache,target=/root/.cargo/registry --mount=type=cache,target=/root/.cargo/git --mount=type=cache,target=/opt/foundry/target \
+    source $HOME/.profile && cargo build --release --features cast/aws-kms,forge/aws-kms \
```
bulenttastan, line 20: Instead of building the Foundry, can we just run the suggested command on their [website](https://getfoundry.sh)? Otherwise, it takes so long on Docker to build it.

RomanHodulak: Does not work with musl libc. The build setup is copied from their [dockerfile](https://github.com/foundry-rs/foundry/blob/master/Dockerfile). It does not seem that they build & push it anywhere public though, so for now we keep it here.

docker/Dockerfile.geth
RomanHodulak: My thoughts exactly, but it seemed to work with libclang even rocksDB had no complains so that makes me happy.

docker/Dockerfile.optimism
RomanHodulak: That's a good point, TBH this line is probably useless at this point

We could probably remove it. Currently, it's simply copied from https://github.com/foundry-rs/foundry/blob/master/Dockerfile with no changes, so at least it's clear that it is replaceable by just copy pasting it without any custom changes. So I decided to keep it as is.

docker/Dockerfile.geth
```diff
@@ -0,0 +1,47 @@
...
+# Copy built binary from build image
+COPY --from=build /volume/build/bin/geth /volume/geth
+
+# Set entrypoint to geth binary
+ENTRYPOINT ["/volume/geth",
+    # Ephemeral proof-of-authority network with a pre-funded developer account,
+    # with automatic mining when there are pending transactions.
+    "--dev",
+    "--datadir",
+    "./l1_datadir",
```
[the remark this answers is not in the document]
RomanHodulak: It would be ideal to have a static faucet address. Having pregenerated keystore might also help with making `geth` restart-able as hinted by https://geth.ethereum.org/docs/developers/dapp-developer/dev-mode#reusing-accounts

docker/prefund.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/sh
+# Entrypoint of the op-batcher docker container
+
+set -eu
+. /volume/.env
+L1_RPC_URL="http://127.0.0.1:58138"
+FROM_WALLET=$(ls l1_datadir/keystore | grep -o '.\{40\}$')
+FACTORY_DEPLOYER_ADDRESS="0x3fAB184622Dc19b6109349B94811493BF2a45362"
+HEARTBEAT_ADDRESS="0x88f9b82462f6c4bf4a0fb15e5c3971559a316e7f"
```
bulenttastan, line 9: We don't need this and the `curl` call below any more.

bulenttastan (approved):
Let's move 👏

For anyone looking to follow along the Docker process, it is easier to compare with the integration test. It replicates the same steps.

PR #277, opened 2025-02-18
feat: Implement payload queries

RomanHodulak:
### Description
This PR improves the implementation of payloads that cause the `op-node` to stall the blockchain indefinitely.

This happens when one payload is requested more than once before starting to build a new block. In that case the payload cannot be retrieved because it is dropped immediately when fetched.

Not only is this a side-effect of a method that should behave idempotently. The payloads are also kept in-memory and inside `StateActor`, making it coupled with implementation and impossible to inject database backed implementation.

### Changes
* Add support for historical fetching of payloads
* Add support for fetching the same payloads repeatedly
* Decouple `StateActor` from payload query implementation
* Declare `PayloadQueries` and inject it into `StateActor`
* Implement `PayloadQueries` in-memory
* Implement `PayloadQueries` using `rocksdb`

### Testing
* Add unit test scenario of getting one payload repeatedly
* Add unit test scenario of getting an older payload

### Notes
This PR saves memory because it creates payload responses from blocks and removes their caching inside `StateActor`. Without this change, the memory would slowly increase over time indefinitely, even when using the database backed implementation.


moved/src/payload/read.rs
```diff
@@ -0,0 +1,21 @@
...
+pub trait PayloadQueries {
+    type Err: Debug;
+    type Storage;
+
+    fn by_hash(
+        &self,
+        storage: &Self::Storage,
+        block_hash: B256,
+    ) -> Result<Option<PayloadResponse>, Self::Err>;
+    fn by_id(
```
bulenttastan, line 16: nit: new line space above

moved/src/state_actor.rs
```diff
@@ -1188,4 +1234,157 @@ mod tests {
...
+        let expected_payload = rx.blocking_recv().unwrap();
+
+        let (tx, rx) = oneshot::channel();
+
+        state_actor.handle_query(Query::GetPayload {
+            id: payload_id,
+            response_channel: tx,
+        });
+
+        let actual_payload = rx.blocking_recv().unwrap();
```
bulenttastan, line 1290: nit: maybe we should call `actual_payload` `same_payload` instead

moved/src/state_actor.rs
```diff
@@ -1188,4 +1234,157 @@ mod tests {
 
         assert_eq!(actual_sender_balance, expected_sender_balance);
     }
+
+    #[test]
+    fn test_one_payload_can_be_fetched_repeatedly() {
+        let to = Address::new(hex!("44223344556677889900ffeeaabbccddee111111"));
```
bulenttastan, line 1240: We can make this transaction creation a test helper method.

moved/src/state_actor.rs
```diff
@@ -1188,4 +1234,157 @@ mod tests {
...
+        });
+
+        let (tx, rx) = oneshot::channel();
+
+        state_actor.handle_query(Query::GetPayload {
+            id: payload_id,
+            response_channel: tx,
+        });
+
+        let actual_payload = rx.blocking_recv().unwrap();
```
bulenttastan, line 1386: Instead, we can call it `old_payload` or `same_payload`

moved/src/state_actor.rs
RomanHodulak: Thank you for the suggestion, it's a good idea to name it same_payload to signify that it is expected to be the same.

However, I'd like to keep the naming as-is because it marks which one of these payloads should match which.

If the test fails, we should assume that expected payload is correct and the actual is the one that needs fixing.

moved/src/state_actor.rs
RomanHodulak: True, I considered it, but ultimately I still have a reason to stick with this naming as stated in the other comment.

RomanHodulak:
Thanks for running the integration test, good to know that it passes.

I appreciate your approval.

PR #280, opened 2025-02-21
Feat(server): Health check response on GET requests

bulenttastan:
### Description
Load balancers make health check requests to include VMs in the backend. The requests use GET method. This PR responds to any GET request with 200 success status and `healthy` response.

### Changes
- Shortcut server response if it's a GET request with `healthy` string response

### Testing
✓ Tested on load balancer and the VM got included after this change

### DIFF
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -34,7 +34,7 @@ use {
     warp::{
         hyper::{body::Bytes, Body, Response},
         path::FullPath,
-        Filter, Rejection,
+        Filter, Rejection, Reply,
     },
     warp_reverse_proxy::{
         extract_request_data_filter, proxy_to_and_forward_response, Headers, Method,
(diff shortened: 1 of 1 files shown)


server/src/lib.rs
```diff
@@ -260,6 +260,11 @@ async fn mirror(
 ) -> Result<warp::reply::Response, Rejection> {
     let (path, query, method, headers, body) = request;
 
+    // Handle load balancer health check with a simple response
+    if method == Method::GET {
+        return Ok("healthy".into_response());
```
RomanHodulak, line 265: Is there a dependency on this endpoint returning "healthy" string? Otherwise we can simply return successful no content response.

This might do the trick
```rust
Response::builder()
    .status(StatusCode::NO_CONTENT)
```

RomanHodulak (approved):

server/src/lib.rs
bulenttastan: `NO_CONTENT` works better. `healthy` is just industry standard, but no need for any specific string.

ilchu (approved):

PR #281, opened 2025-02-23
build: Remove optimism submodule

RomanHodulak:
### Description
Using submodules is unneccessary and brings in complexity and messes up the staging area.

It is also inconsistent with other dependencies like `geth` that are cloned without submodule.

This PR gets rid of the submodule setup while keeping everything adjusted to cloning and makes it all consistent.

### Changes
- Remove submodules from the repository
- Adjust readme instructions for the cloning
- Ignore optimism directory in git to avoid adding it and including it in git status

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,17 +6,17 @@ A Move VM execution layer for OP Stack.
 
 Make sure you have `go` installed on your system. Other dependencies include [foundry](http://getfoundry.sh/) for smart contract interaction and [jq](https://jqlang.github.io/jq/) being called indirectly by Optimism itself.
 
-Optimism monorepo is pulled in as a submodule of this repo. The submodule is used to compile and deploy Optimism contracts.
-If you don't see it inside the test folder, try bringing in the Optimism submodule manually
+First, clone the Optimism monorepo. The repo is used to compile and deploy Optimism contracts.
 
 ```bash
-git submodule update --init --recursive
+git clone https://github.com/ethereum-optimism/optimism server/src/tests/optimism
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -6,17 +6,17 @@ A Move VM execution layer for OP Stack.
 
 Make sure you have `go` installed on your system. Other dependencies include [foundry](http://getfoundry.sh/) for smart contract interaction and [jq](https://jqlang.github.io/jq/) being called indirectly by Optimism itself.
 
-Optimism monorepo is pulled in as a submodule of this repo. The submodule is used to compile and deploy Optimism contracts.
-If you don't see it inside the test folder, try bringing in the Optimism submodule manually
+First, clone the Optimism monorepo. The repo is used to compile and deploy Optimism contracts.
 
 ```bash
-git submodule update --init --recursive
+git clone https://github.com/ethereum-optimism/optimism server/src/tests/optimism
```
bulenttastan, line 12: This is exactly what the submodule does, in fact. Not sure which one the team prefers though. So need more opinion on this. @ilchu what do you think?

ilchu: Well keeping it as a submodule did make it a nuisance in the file history in git, for one thing. I do feel that the submodule way could serve its purpose if we try to keep up with the official latest optimism repo, but that could be a pain still if things break. So I do agree that it's better to just clone it for now.

ilchu (approved):

bulenttastan (approved):

PR #282, opened 2025-02-24
refactor: Separate orchestration logic from `moved` into `moved-app` crate

RomanHodulak:
### Description

Insert description here

### Changes

- Change 1
- Change 2


app/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+pub(crate) mod input;
+
+mod actor;
+
+pub use {actor::*, input::*};
```
bulenttastan, line 5: nit: I believe we keep `use` at the top

bulenttastan (approved):
Looks good. I like the crate name `moved-app`

PR #286, opened 2025-02-28
refactor(server): Split per-feature dependencies using modules

RomanHodulak:
### Description
Instead of having one module split using compiler flags, we split it into two distinct modules and then import one of them using the same flag.

This makes the individual modules more compact and easy to work with. 

Lays some groundwork for the upcoming LMDB integration.

### Changes
- Split `storage` and `in_memory` dependency stack into two modules.

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt


bulenttastan (approved):
Nice, to try another DB type.

PR #287, opened 2025-02-28
feat(storage): Implement `BlockRepository` using `heed`

RomanHodulak:
### Description
Adds the first implementation using `heed`.

The crate `heed` defines the rust bindings for LMDB.

### Changes
- Create crate `storage-heed`
- Implement `BlockRepository` using `heed`

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt


Cargo.toml
```diff
@@ -13,6 +13,7 @@ members = [
     "shared",
     "state",
     "storage",
+    "storage-heed",
```
bulenttastan, line 16: I think we should keep these package inside `storage`.

RomanHodulak: I think so too.

storage/heed/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "moved-storage-heed"
+description = "Moved node integration with LMDB persistent storage engine via Heed bindings"
+version.workspace = true
+edition.workspace = true
+
+[dependencies]
+moved-blockchain.workspace = true
+moved-shared.workspace = true
+heed.workspace = true
```
bulenttastan, line 10: nit: alphabetically should be the first in dependencies.

bulenttastan (approved):

PR #294, opened 2025-03-01
feat(server): Integrate with LMDB persistent backing storage engine using `heed`

RomanHodulak:
### Description
Follows-up on previous PRs about LMDB implementation and integrates it with the server into an executable binary.

### Changes
- Add `server` configuration with `heed` storage
- Add transaction commit per database call
- Create LMDB environment with 1 TiB total size limit

### Testing
Long running `docker compose down && docker compose build && docker compose up -d`

Collecting statistics using `pidstat`

### Notes
To run with LMDB:
```
cargo run --bin op-move --features storage-lmdb
```

### DIFF
--- a/docker/Dockerfile.op-move
+++ b/docker/Dockerfile.op-move
@@ -28,7 +28,7 @@ RUN --mount=type=cache,target=/root/.cargo/registry \
 # RocksDB breaks the build with libclang.so not found https://github.com/apache/skywalking/issues/10439
 # Aptos-core breaks the build with `disable_lifo_slot` not found https://github.com/aptos-labs/aptos-core/issues/5655 \
     RUSTFLAGS="-C target-feature=-crt-static --cfg tokio_unstable" \
-    cargo build --bin op-move --release --features storage \
+    cargo build --bin op-move --release --features storage-lmdb \
     && mv target/release/op-move /volume/op-move
 
 # Switch to clean image
--- a/server/src/dependency/heed.rs
+++ b/server/src/dependency/heed.rs
@@ -1,0 +1,189 @@
+use {
+    crate::dependency::shared::*,
+    moved_blockchain::{
+        block::{BaseGasFee, BlockHash, BlockRepository, MovedBlockHash},
+        payload::NewPayloadId,
+    },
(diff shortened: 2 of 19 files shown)


docker/Dockerfile.op-move
```diff
@@ -28,7 +28,7 @@ RUN --mount=type=cache,target=/root/.cargo/registry \
 # RocksDB breaks the build with libclang.so not found https://github.com/apache/skywalking/issues/10439
 # Aptos-core breaks the build with `disable_lifo_slot` not found https://github.com/aptos-labs/aptos-core/issues/5655 \
     RUSTFLAGS="-C target-feature=-crt-static --cfg tokio_unstable" \
-    cargo build --bin op-move --release --features storage \
+    cargo build --bin op-move --release --features storage-lmdb \
```
bulenttastan, line 31: We should organize these feature names and package names later. Rocks DB uses just `storage` as the feature name, but LMDB has the full name. Also the packages are called with `heed` not LMDB.

server/src/dependency/heed.rs
```diff
@@ -0,0 +1,189 @@
...
+    if std::fs::exists(path).unwrap() {
+        std::fs::remove_dir_all(path)
+            .expect("Removing non-empty database directory should succeed");
+    }
+    std::fs::create_dir(path).unwrap();
+
+    let env = unsafe {
+        EnvOpenOptions::new()
+            .max_dbs(moved_storage_heed::DATABASES.len() as u32)
+            .map_size(1024 * 1024 * 1024 * 1024) // 1 TiB
```
bulenttastan, line 149: Does it pre-allocate the data file, meaning do we need a server over 1 TiB storage?

bulenttastan (approved):

server/src/dependency/heed.rs
RomanHodulak: No, this is an upper limit of the database directory size on disk.

Internally, it means the size of the virtual address space.

docker/Dockerfile.op-move
RomanHodulak: It's true that Rocks DB should also have the feature flag name extended to be consistent and transparent on the meaning of these flags.

My thoughts on the naming were: Looking at it from the user's perspective, I care about using `lmdb`, rather than the fact that it is ported through `heed`.

PR #298, opened 2025-03-04
feat(storage-heed): Wrap database in a custom type

RomanHodulak:
### Description
Having a custom type for a database gives us a better modeling opportunities. It allows us to only permit behavior that we expect to be using. Thanks to zero-cost abstractions, this comes at no performance cost.

What's more it gives us a better control and transparency over which calls are being made. When all the calls to database go through our type, we can extend the calls to always include our custom behavior. For example, we can log every call or place a debugger breakpoint in there. This helps to debug a performance issue, for example.

It also reduces code duplication of redundant database opening calls and error handling.

### Changes
- Add wrapper type for `heed::Database`
- Add extension traits for opening known databases
- Implement extension traits for `heed::Env`

### Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt
:green_circle: `docker-compose down && docker-compose build && docker-compose up -d`

### DIFF
--- a/storage/heed/src/all.rs
+++ b/storage/heed/src/all.rs
@@ -1,4 +1,7 @@
-use crate::{block, payload, receipt, state, transaction, trie};
+use {
+    crate::{block, payload, receipt, state, transaction, trie},
+    heed::{types::LazyDecode, BytesDecode, BytesEncode, RoTxn, RwTxn},
+};
 
 pub const DATABASES: [&str; 9] = [
     block::DB,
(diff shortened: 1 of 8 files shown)


storage/heed/src/all.rs
```diff
@@ -12,6 +15,40 @@ pub const DATABASES: [&str; 9] = [
     payload::DB,
 ];
 
+#[derive(Debug)]
+pub struct HeedDb<KC, DC>(pub heed::Database<KC, DC>);
+
+impl<KC, DC> HeedDb<KC, DC> {
```
bulenttastan, line 21: nit: why not `K` and `D` instead of `KC` and `DC`

storage/heed/src/all.rs
```diff
@@ -12,6 +15,40 @@ pub const DATABASES: [&str; 9] = [
...
+    pub fn put<'a>(
+        &self,
+        txn: &mut RwTxn,
+        key: &'a KC::EItem,
+        data: &'a DC::EItem,
```
bulenttastan, line 26: nit: maybe we should call it `value` instead of `data`, like key -> value

bulenttastan (approved):

PR #299, opened 2025-03-05
Fix: Simplify cloud logs and get eth_estimateGas tx input data

bulenttastan:
### Description
A few fixes combined.

### Changes
- Cloud logs are better when they are in a single line
- `eth_estimateGas` transaction input can come from either `input` or `data`
- Serve the Devnet with the same key file so the faucet address doesn't change

### Testing
✓ Docker works
✓ Hardhat is able to deploy by calling `eth_estimateGas`
✓ Cloud logs come through one request at a time, instead of each line in the request JSON being split into a separate log

### DIFF
--- a/docker/geth.sh
+++ b/docker/geth.sh
@@ -6,10 +6,12 @@
 # -x Print out command arguments during execution
 set -eux
 L1_DATADIR="./l1_datadir"
+KEYFILE="89d740330e773e42edf98bba1d8d1d6c545d78a6"
 
 # In case of restart, data dir will not be empty
 # Wipe it to avoid an attempt to restore the state which is not persisted in `--dev` mode
-rm -rf "${L1_DATADIR}" && mkdir -p "${L1_DATADIR}"
+rm -rf "${L1_DATADIR}" && mkdir -p "${L1_DATADIR}" && mkdir -p "${L1_DATADIR}/keystore"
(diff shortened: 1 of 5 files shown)


docker/geth.sh
```diff
@@ -6,10 +6,12 @@
 # -x Print out command arguments during execution
 set -eux
 L1_DATADIR="./l1_datadir"
+KEYFILE="89d740330e773e42edf98bba1d8d1d6c545d78a6"
 
 # In case of restart, data dir will not be empty
 # Wipe it to avoid an attempt to restore the state which is not persisted in `--dev` mode
-rm -rf "${L1_DATADIR}" && mkdir -p "${L1_DATADIR}"
+rm -rf "${L1_DATADIR}" && mkdir -p "${L1_DATADIR}" && mkdir -p "${L1_DATADIR}/keystore"
+echo '{"address":"89d740330e773e42edf98bba1d8d1d6c545d78a6","crypto":{"cipher":"aes-128-ctr","ciphertext":"755bbdc3f7bed78a5434fcce4dc118795dd049b0538084c3ce12a91035b32c60","cipherparams":{"iv":"22e00eec04c8d117ad12c0b2924f6a6b"},"kdf":"scrypt","kdfparams":{"dklen":32,"n":4096,"p":6,"r":8,"salt":"cab7899fd54ef0164f572aa41b3941d366ccd38ec624d06ee4e8ee088ef4da3d"},"mac":"0b0a2137144d1efdea0083a22e265300a11bf0d3e5b55285f4ce366a8eb1f8bf"},"id":"52f30552-f758-4cd0-97d4-7f957b51ef30","version":3}' > "${L1_DATADIR}/keystore/${KEYFILE}"
```
RomanHodulak, line 14: The dev account address is defined multiple times. Once in `address` and once in `KEYFILE`.

RomanHodulak (approved):
Good to have this committed.

docker/geth.sh
bulenttastan: I'm gonna rename the keyfile the way `geth` names it by default with date, as:
`UTC--2025-01-01T00-00-00.000000000Z--89d740330e773e42edf98bba1d8d1d6c545d78a6`.

bulenttastan: I changed the way keystore file is created and made it a separate bash file. @RomanHodulak  What do you think of this way in https://github.com/MovedNetwork/op-move/pull/299/commits/34c8606902ff256aaf720806736c8b2a3b5dd9e8?

server/src/lib.rs
```diff
@@ -318,7 +315,7 @@ async fn mirror(
         op_move_response: &op_move_response,
         port,
     };
-    println!("{}", serde_json::to_string_pretty(&log).unwrap());
+    println!("{}", serde_json::to_string(&log).unwrap());
```
ilchu, line 318: Should we consider moving to something like [`tracing`](https://docs.rs/tracing/latest/tracing/) eventually? Then we could have different sinks for jsonified or prettified log output, for example. Among other observability knick-knacks.

ilchu (approved):

Issue #296, opened 2025-03-02
Optimize database entity encoding


RomanHodulak:
The database size follows an exponential trend and is growing at an unsustainable rate.

![Image](https://github.com/user-attachments/assets/73614b6a-4215-41f1-ad1b-e63a28e97191)

This is largely linked to inefficient encoding. All the entities are encoded using JSON. For example, a simple hash has a length of 32 bytes but with JSON requires 64 bytes worth of hex characters + 2 bytes of `0x` prefix + 2 bytes of double quotes + 1 byte of colon character + 2 bytes of double quotes + the name of the field + up to 1 byte of comma character.

So for `"withdrawals":"0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",` you need 83 bytes, which is almost triple of what it needs to store. This field always has one constant value, so there is effectively 0 bytes worth of information.

A clever encoding would make use of that assumption. It would also not encode field names and possibly used more techniques like variably sized integers.

However, the entities from the moved-blockchain domain contain objects that are written specifically for JSON and make use of certain serde flags like "skip_serializing_if" which breaks many binary formats such as `bincode` or `bcs`. These formats expect certain amount of fields in a struct and cannot usually deserialize what they previously serialized when using these structs.

It would also be good to have separate types for any database migration / versioning purposes.

https://github.com/MovedNetwork/op-move/blob/14306a86a25ca594f01055700f7b116a3c0deda5/storage/heed/src/generic.rs#L34-L57

---
Metrics from the Cloud VM with configuration: 16 CPU + 124 GB Memory + 2 TB Hard Drive

<img width="784" alt="Image" src="https://github.com/user-attachments/assets/4bcef915-c0c4-42ab-bef8-68194fcf389a" />

PR #306, opened 2025-03-12
perf: Divide scaling laws by `n`

RomanHodulak:
# About
Implementation of an optimization solution to a fatal performance issue, preventing the `op-move` node from running persistently without exhausting resources.

# Motivation
The usage of resources by the op-node is scaling at an unsustainable rate. So much so that even running release optimized code completely in-memory lasts for only about 24 hours.

This effectively prevents the devnet from being usable. The whole state, including every deployed user contract, needs to be wiped out and reset from genesis once per day. It also negatively affects availability, reliability and latency and completely blocks testnet launch.

# Problem
After analysis of the database write calls, the problem was clearly identified inside the EVM native extension as the usage of the EVM storage trie writes. Every EVM account has a storage trie serialized into a single Move resource. There is exactly one merkle trie node for one Move resource.

Because the merkle trie needs to keep every historical version of every leaf node, having it completely replaced means that for every update of the storage trie, the whole node is written again as a new one with the addition of the new writes.

Because for every block n we need to perform n writes and a constant amount of serialization and deserialization, this leads to the effect of **quadratic** memory complexity and **linear** time complexity.

# Solution
Because the EVM account storage is represented by a trie, we can conclude that we only need to solve the way that it is stored and loaded. Serializing an entire merkle trie into a single trie node effectively nullifies the trie capacity of recording changes only by appending new entries and new versions of existing entries.

Furthermore, the EVM extension is implemented as a native function extension. This means that we are not limited to the capacity of the Move embedded environment and have the freedom to implement the storage layer in whichever way we want. This is because the native functions always end up calling the functions on the host environment.

Therefore, instead of using Move resource, we can represent the storage directly as a trie, just like the Move state. Having them aside represented as a trie aligns directly with the Ethereum design and does not pose the same problems as circumventing this design by fitting all the data in one merkle node.

### Before
measure1
We can observe that
* CPU usage: linear
* Disk usage: quadratic
* Disk writes: linear

### After
measure2
We can observe that
* CPU usage: constant
* Disk usage: linear
* Disk writes: constant

### Discussion
We can observe that the scaling laws were divided by `n`. This is a fantastic result that brings in an enourmous improvement in the node's performance. We can also observe that the memory usage has massively dropped to being at around 0.3%, which should be around 300 MB.

### Conclusion
In this way the node is run sustainably as long as there is enough disk space. During these measurements, the node takes around a gigabyte per 24 hours. This can be further optimized by lowering the constant rate factor posed by the encoding. There is a relatively large room for improvement as discussed in #296.

# Changes
- Add `StorageTrieDb` trait that creates `eth_trie::DB` implementation for the storage trie.
- Implement `StorageTrieDb`
  - in-memory using standard library
  - LMDB via `heed`
  - RocksDB via `rocksdb`
- Generate `Changes` representing commit to be applied to storage tries.
- Apply storage tries `Changes` on new blockchain state.

# Testing
:green_circle: Build
:green_circle: Test   
:green_circle: Clippy
:green_circle: Rustfmt
:green_circle: `docker compose down && docker compose build && docker compose up -d` and collecting the stats.


evm-ext/src/state/storage.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    fn flush(&self) -> result::Result<(), Self::Error> {
+        Ok(self.0.flush()?)
+    }
+}
+
+impl StorageTrieDb for () {
+    fn db(&self, _: Address) -> Arc<BoxedTrieDb> {
+        todo!()
```
bulenttastan, line 349: is this a real `// TODO` or unreachable?

blockchain/src/state/read.rs
```diff
@@ -158,23 +165,26 @@ pub fn proof_from_trie_and_resolver(
...
+                        key: (*index).into(),
                         value,
                         proof: proof.into_iter().map(Into::into).collect(),
-                    }
+                    })
                 })
             })
-            .collect()
+            .collect::<Result<_, _>>()
+            .unwrap()
```
bulenttastan, line 187: Maybe we shouldn't `.unwrap()`

bulenttastan (commented):
Great work 👏
I don't see any major issues, but I'll let @birchmd to take a look before approving.

blockchain/src/state/read.rs
RomanHodulak: Definitely.

At this point this function does not return result and a lot of the errors are ignored by returning `None`. But it's proving to cost some confusion when debugging as the function reports as "not found." while in some cases it really means "error."

So without error handling I'd say it's better to unwrap. While it does cause a panic, it makes it very clear where the error happened.

I think implementing proper result should be a part of a broader issue focused on error handling.

birchmd (approved):
Great work on this PR! Since it is such an important fix and because it is already quite a large PR, I think we should go ahead and merge it as-is. But, I think there should be two follow-up PRs to address some minor problems. First, I think I identified a place where were are prematurely writing to the underlying DB and this should be fixed. And second I think it is important to revisit error handling around Eth Trie since it is now fallible.

blockchain/src/state/read.rs
ilchu: I agree on the error handling needing to be separate, as there are a lot of things to revamp generally and I prefer PRs being semantic. Bulent's comment would also apply to e.g. panics [here](https://github.com/MovedNetwork/op-move/pull/306/files#diff-e1f6436696c104b1bc70e72f50808a691759740596e20125b88d777e1eb8f4feR580), though.

evm-ext/src/native_evm_context.rs
```diff
@@ -33,84 +32,49 @@ pub struct HeaderForExecution {
...
-    // size of a single resource, a cache will still be useful since reconstructing
-    // the storage from all the resource pieces will also be non-trivial.
-    //
-    // The cache must be wrapped in a type that allows interior mutability
-    // because the `DatabaseRef` interface uses immutable references. Since this
-    // cache is only used for a single transaction execution it is unlikely that
-    // it will ever need to multi-threaded access, so a thread-unsafe type like
-    // `RefCell` would be sufficient. But at the same time, I don't think the overhead
-    // of `RwLock` is that large, so I think it is ok to use.
-    storage_cache: RwLock<HashMap<Address, trie_types::AccountStorage>>,
```
ilchu, line 75: Just a side thought: any chance we would still benefit from having a cache of some sorts?

evm-ext/src/state/storage.rs
```diff
@@ -0,0 +1,388 @@
...
+};
+
+/// [`result::Result`] with its `Err` variant set to [`Error`].
+pub type Result<T> = result::Result<T, Error>;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("{0}")]
+    EthTrie(#[from] TrieError),
+    #[error("{0}")]
```
ilchu, line 23: Maybe `#[error(transparent)]` here?

execution/src/transaction.rs
```diff
@@ -234,6 +234,27 @@ impl NormalizedExtendedTxEnvelope {
     }
 }
 
+type MoveChanges = ChangeSet;
+type EvmChanges = moved_evm_ext::state::StorageTriesChanges;
+
+#[derive(Debug, Clone)]
+pub struct Changes {
+    pub r#move: MoveChanges,
```
ilchu, line 242: I remember there was some discussion before, but how about just calling it `moved`? So that we don't have to make it a raw ident. Though it's a nit, of course.

evm-ext/src/native_impl.rs
```diff
@@ -165,7 +171,7 @@ fn evm_transact_inner(
 
     // Commit the changes to the DB so that future Move transactions using
     // the same session will see them.
-    evm_native_ctx.db.commit(outcome.state);
+    db.commit(outcome.state);
```
ilchu, line 175: Correct me if I'm wrong, but do we still need to commit changes to the db at the end? As now it's not part of context, it's not persisted across invocations of this fn. And from what I see in `commit()` source code, it doesn't mutate the actual db that was passed into it (and thus storage tree or resolver with it), but only modifies its inner mappings for accounts and contracts.

RomanHodulak: You're right, that has no effect now.

execution/src/transaction.rs
RomanHodulak: I think the best name is `move_vm` as it contains the name of the embedded environment where the changes were generated.

evm-ext/src/state/storage.rs
RomanHodulak: Maybe yes

evm-ext/src/native_evm_context.rs
RomanHodulak: I don't think so. This was closely tied to the Move resource serialization scheme. In that design, you had to deserialize the entire trie from a single byte array and ended up with a trie structure that's all in-memory. This approach tried to optimize the deserialization logic overhead, but in the end it was not enough.

evm-ext/src/native_evm_context.rs
```diff
@@ -148,19 +112,9 @@ impl<'a> DatabaseRef for ResolverBackedDB<'a> {
...
-        if let Some(storage) = cache_lock.get(&address) {
-            return Ok(storage.get(index));
-        }
-
-        let storage = self.storage_for(&address)?;
-        let value = storage.get(index);
-        cache_lock.insert(address, storage);
-        Ok(value)
+        let storage = self.storage_trie.for_account(&address).unwrap();
+        let value = storage.get(&index).unwrap();
```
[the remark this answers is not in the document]
RomanHodulak: Yes that would be much better.

execution/src/transaction.rs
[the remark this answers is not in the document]
RomanHodulak: I think the name `move_vm` is more appropriate as it is the name of the environment where the changes were generated.

evm-ext/src/state_changes.rs
```diff
@@ -138,44 +157,19 @@ fn add_account_changes(
...
-    let storage_bytes = storage.serialize();
-    let op = if is_created {
-        Op::New(storage_bytes.into())
-    } else {
-        Op::Modify(storage_bytes.into())
-    };
-    result
-        .add_resource_op(storage_tag, op)
-        .expect("Resource cannot already exist in result");
+    let storage_changes = storage.commit().unwrap();
```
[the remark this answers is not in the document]
RomanHodulak: Yes, that has slipped through the cracks, so thanks I really appreciate your feedback.

Solved.

birchmd (approved):
Let's ship it 🚀

PR #307, opened 2025-03-13
Feat(integration-test): Test for ERC-20 deposits via the standard bridge

birchmd:
### Description
This PR extends the end-to-end integration test to include bridging ERC-20 tokens. The basic design is to use the `OptimismMintableERC20Factory` which is part of the OP stack to allow the standard bridge contract work as intended. This means the bridged ERC-20 tokens only exist inside the EVM, but I think this is acceptable. We add a new Move interface to allow Move contracts to interact with the bridged ERC-20 tokens.

### Changes
- New Move interface in our Framework to access ERC-20 token deployed to the EVM.
- ERC-20 deposit flow added to integration test

### Testing
- Integration test

### Notes
- The ERC-20 Move API is obviously incomplete; this will be finished in a follow-up PR.
- The withdrawal of ERC-20 tokens will also be handled in a separate PR.


genesis-builder/src/main.rs
```diff
@@ -55,6 +55,7 @@ static APTOS_PACKAGE_PATHS: Lazy<Vec<PathBuf>> = Lazy::new(|| {
         MOVED_FRAMEWORK_DIR.join("eth-token"),
         MOVED_FRAMEWORK_DIR.join("evm"),
         MOVED_FRAMEWORK_DIR.join("l2"),
+        MOVED_FRAMEWORK_DIR.join("erc20-api"),
```
bulenttastan, line 58: The folder name would be just `erc20`

server/src/tests/integration.rs
```diff
@@ -435,13 +436,47 @@ fn run_op() -> Result<(Child, Child, Child)> {
...
+
+    deposit_erc20_to_l2().await?;
+
     withdrawal::withdraw_to_l1().await?;
 
     Ok(())
 }
 
-async fn deposit_to_l2() -> Result<()> {
+async fn deposit_erc20_to_l2() -> Result<()> {
```
bulenttastan, line 448: We probably wanna move this method to `erc20.rs`

bulenttastan, line 448: nit: We can move this method below `deposit_eth_to_l2` to match the order of calls.

server/src/tests/integration.rs
```diff
@@ -435,13 +436,47 @@ fn run_op() -> Result<(Child, Child, Child)> {
...
+    // Deploy ERC-20 token to bridge
+    let l1_address = erc20::deploy_l1_token(&from_wallet, &l1_rpc).await?;
+    // Create corresponding token on L2
+    let l2_address = erc20::deploy_l2_token(&from_wallet, l1_address, L2_RPC_URL).await?;
+    // Perform deposit
+    erc20::deposit_l1_token(
+        &from_wallet,
+        l1_address,
+        l2_address,
+        deposit_amount,
```
bulenttastan, line 463: nit: If we rename `deposit_amount` to `amount` this func call fits into a single line.

bulenttastan (approved):

genesis-builder/framework/erc20-api/Move.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "Erc20Api"
```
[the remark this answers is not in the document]
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/307/commits/82ad18b01634c92e1bde1af47ed72117acb9652b

genesis-builder/src/main.rs
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/307/commits/82ad18b01634c92e1bde1af47ed72117acb9652b

server/src/tests/integration.rs
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/307/commits/3de586aa4b6401bee698efda703acb6428e7379d

server/src/tests/integration.rs
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/307/commits/3de586aa4b6401bee698efda703acb6428e7379d

server/src/tests/integration/erc20.rs
```diff
@@ -0,0 +1,164 @@
...
+        .with_recommended_fillers()
+        .wallet(EthereumWallet::from(from_wallet.to_owned()))
+        .on_http(Url::parse(rpc_url)?);
+
+    let contract = erc20_factory::OptimismMintableERC20Factory::new(factory_address, provider);
+    let receipt = contract
+        .createOptimismMintableERC20(l1_address, NAME.into(), SYMBOL.into())
+        .send()
+        .await
+        .unwrap()
```
ilchu, line 81: Maybe just `?` all the unwraps in this func and the balance one as we're returning `anyhow` result anyways?

server/src/tests/integration/erc20.rs
```diff
@@ -0,0 +1,164 @@
...
+        .await
+        .unwrap();
+    let event = receipt
+        .inner
+        .logs()
+        .iter()
+        .find(|log| log.address() == factory_address)
+        .expect("OptimismMintableERC20Factory emits log");
+    let event = event
+        .log_decode::<erc20_factory::OptimismMintableERC20Factory::StandardL2TokenCreated>()
```
ilchu, line 92: Optimism themselves suggest to rely on the non-legacy `OptimismMintableERC20Created` event instead, but that probably comes right after the first matching find in the iterator. It also has the order of local and remote token switched, but it shouldn't matter, as the name is likely same.

server/src/tests/integration/erc20.rs
```diff
@@ -0,0 +1,164 @@
...
+) -> Result<()> {
+    let provider = ProviderBuilder::new()
+        .with_recommended_fillers()
+        .wallet(EthereumWallet::from(from_wallet.to_owned()))
+        .on_http(Url::parse(rpc_url)?);
+
+    let bridge_address = Address::from_str(&get_deployed_address("L1StandardBridgeProxy")?)?;
+    let bridge_contract = bridge::L1StandardBridge::new(bridge_address, provider);
+    let receipt = bridge_contract
+        .depositERC20(l1_address, l2_address, amount, 21_000, Default::default())
```
ilchu, line 115: I'm a bit uncomfortable about hardcoding gas limits. Are they guaranteed to be deterministic? Or maybe better to have some surplus for leeway, or even issue an estimate gas call beforehand.

server/src/tests/integration.rs
```diff
@@ -463,6 +467,30 @@ async fn deposit_to_l2() -> Result<()> {
     Ok(())
 }
 
+async fn deposit_erc20_to_l2() -> Result<()> {
+    let l1_rpc = var("L1_RPC_URL").unwrap();
```
ilchu, line 471: The older code was mostly using `expect()` for env vars, but it's a nit.

ilchu (approved):

server/src/tests/integration/erc20.rs
birchmd: I was unwrapping just for debugging purposes since then the specific line number shows up in the error message, whereas `?` will propagate the error back to the test runner. We could mitigate this issue by using the `context` machinery that `anyhow` offers, but for now I'll just put in the question marks.

Done in https://github.com/MovedNetwork/op-move/pull/307/commits/5ece865158ea1c6db900db6338c90bc90cfd9c7b

server/src/tests/integration.rs
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/307/commits/5ece865158ea1c6db900db6338c90bc90cfd9c7b

server/src/tests/integration/erc20.rs
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/307/commits/e3bc6493e2d9f45f19f0a26b3865c074f94bde04

server/src/tests/integration/erc20.rs
bulenttastan: I think the surplus is acceptable here. SDKs call estimate gas before making calls, so that flow is being tested by the SDKs.

server/src/tests/integration/erc20.rs
bulenttastan: Good call 🙌

bulenttastan (approved):

PR #311, opened 2025-03-17
perf: Generate payload ID before starting block build

RomanHodulak:
# About
Unblocks forkchoice API handler from waiting on generated payload ID.

From a general point of view, unblocks any API handler that sends commands from waiting on response as no command should have a response to begin with. This PR makes it so that this invariant holds true.

# Motivation
The application layer receives two different kinds of messages. 

* **Queries:** Immutable and always returns value
* **Commands:** Mutable and never returns value

Working with this assumption, the commands are always handled asynchronously in a background task and do not block the thread that dispatched it - the RPC API method handler. In this way the method handler can return without waiting for the task to finish as there is no reason to wait.

This assumption, however, is broken by one command, which is the `StartBlockBuild`.

# Problem
When sending `StartBlockBuild` to the asynchronous queue for processing, it requires us to wait for its response using a `oneshot` channel. This effectively makes the background task block as if it was foreground.

What's more the request is not being processed immediately as there might be older unprocessed messages in the queue. This makes the API wait for the whole message queue to be processed before it can return a response.

# Solution
Instead of making the command handler responsible for generating the payload ID, make it receive the payload ID and move the responsibility to the layer above. The API handler generates payload ID and returns it in its response. Now there is no need for the command to have a response or for it to block the API handler.

To posses the ability to generate payload IDs, the API delegates this to an abstract dependency. It works the same way as when called from the `StateActor`. It is now simply a dependency of the API handler instead.

# Changes
- Remove `NewPayloadId` dependency from `StateActor`
- Add `NewPayloadId` dependency to `forkchoiceUpdated` request handler
- Remove `oneshot` response sender from `Command::StartBlockBuild`
- Add payload ID to `Command::StartBlockBuild`
- Remove waiting on `oneshot` response receiver when sending `Command::StartBlockBuild` by directly returning the generated ID

# Testing
:green_circle: Build
:green_circle: Test
:green_circle: Clippy
:green_circle: Rustfmt
:green_circle: `docker compose down && docker compose build && docker compose up -d`

### DIFF
--- a/api/src/methods/forkchoice_updated.rs
+++ b/api/src/methods/forkchoice_updated.rs
@@ -8,16 +8,24 @@ use {
             PayloadStatusV1, Status,
         },
     },
-    moved_app::{Command, StateMessage},
-    tokio::sync::{mpsc, oneshot},
+    moved_app::{Command, Payload, StateMessage, ToPayloadIdInput},
+    moved_blockchain::payload::NewPayloadId,
+    tokio::sync::mpsc,
 };
 
 pub async fn execute_v3(
     request: serde_json::Value,
     state_channel: mpsc::Sender<StateMessage>,
+    payload_id: &impl NewPayloadId,
 ) -> Result<serde_json::Value, JsonRpcError> {
     let (forkchoice_state, payload_attributes) = parse_params_v3(request)?;
-    let response = inner_execute_v3(forkchoice_state, payload_attributes, state_channel).await?;
+    let response = inner_execute_v3(
+        forkchoice_state,
+        payload_attributes,
+        state_channel,
+        payload_id,
+    )
+    .await?;
     Ok(serde_json::to_value(response).expect("Must be able to JSON-serialize response"))
 }
 
@@ -52,6 +60,7 @@ async fn inner_execute_v3(
     forkchoice_state: ForkchoiceStateV1,
     payload_attributes: Option<PayloadAttributesV3>,
     state_channel: mpsc::Sender<StateMessage>,
+    payload_id_generator: &impl NewPayloadId,
 ) -> Result<ForkchoiceUpdatedResponseV1, JsonRpcError> {
     // Spec: https://github.com/ethereum/execution-apis/blob/main/src/engine/cancun.md#specification-1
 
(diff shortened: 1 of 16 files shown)


server/src/dependency/heed.rs
```diff
@@ -52,14 +49,13 @@ pub fn state_query(genesis_config: &GenesisConfig) -> StateQueries {
...
     A: State,
-    B: NewPayloadId,
     C: BlockHash,
     D: BlockRepository<Storage = SharedStorage>,
     E: BaseGasFee,
     F: CreateL1GasFee,
     G: CreateL2GasFee,
     H: BaseTokenAccounts,
->() -> OnTxBatch<A, B, C, D, E, F, G, H> {
+>() -> OnTxBatch<A, C, D, E, F, G, H> {
```
bulenttastan, line 58: Instead of removing the `B`, we can shift them and go from `A .. G`. Same applies to the ones below.

bulenttastan (commented):
The code looks good. To look into the impact of this change I'll let @birchmd to take a look too.

api/src/methods/forkchoice_updated.rs
```diff
@@ -72,14 +81,17 @@ async fn inner_execute_v3(
 
     // If `payload_attributes` are present then tell state to start producing a new block
     let payload_id = if let Some(attrs) = payload_attributes {
-        let (tx, rx) = oneshot::channel();
+        let payload_attributes = Payload::from(attrs);
+        let payload_id = payload_id.new_payload_id(
```
ilchu, line 85: Now that I think about it, it's a bit confusing to call this struct `payload_id`, when what it does is actually generating ids. Considering it's a trait, one obvious thing would be to call that trait itself `PayloadIder`, but it's ugly. Whereas `PayloadIdGenerator` is a bit of a mouthful. But I believe some more precise naming would be beneficial.

ilchu (commented):
Uniformity in handling commands is always nice. Seems pretty straightforward and even more explicit now, in tests especially.

api/src/methods/forkchoice_updated.rs
birchmd: I agree it should be clear that we passing in something with the capability to create payload ids as opposed to something that is itself already a payload_id.

birchmd (approved):
Nice work! I like that in addition to a performance improvement, this change simplifies the state actor (which has so many capabilities already).

PR #317, opened 2025-03-27
fix(move): do not encode empty args

### LINKED ISSUE #316: Save gas by not encoding empty arguments in EVM functions for Move
As mentioned in https://github.com/MovedNetwork/op-move/pull/315#discussion_r2014095513, spending Move gas for encoding an empty struct is wasteful, as per Solidity ABI spec the function selector is a prefix as-is and no additional arguments are needed. Currently our generated L2 contracts still have empty arg structs, and we can get rid of them.

ilchu:
### Description

Make parameter encoding step more efficient by skipping it for empty args in L2 contracts.

Fixes #316.

### Changes

- Template file change
- New boolean flag in generation logic

### Testing

Old tests pass as they should.

### DIFF
--- a/genesis-builder/src/generate.rs
+++ b/genesis-builder/src/generate.rs
@@ -17,6 +17,7 @@ struct L2Module {
     functions: Vec<L2Function>,
     structs: BTreeMap<String, Vec<L2Input>>,
     has_fungible_asset: bool,
+    has_non_empty_args: bool,
 }
 
 #[derive(Default, Serialize)]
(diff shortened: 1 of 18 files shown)


genesis-builder/src/generate.rs
```diff
@@ -17,6 +17,7 @@ struct L2Module {
     functions: Vec<L2Function>,
     structs: BTreeMap<String, Vec<L2Input>>,
     has_fungible_asset: bool,
+    has_non_empty_args: bool,
```
bulenttastan, line 20: We could name this just `has_args`

PR #318, opened 2025-03-27
Fix(tx): Nonce validation might run out of gas

birchmd:
### Description
This PR fixes a minor error handling issue where it is not an invariant violation for the nonce check to run out of gas. It could be an invariant violation if we enforce a minimum gas amount for transactions. We should do this for sure, but it has to wait until we handle the gas values more broadly. In the meantime it doesn't hurt to code defensively against the nonce check running out of gas.

### Changes
- Out of gas during nonce check is an invalid transaction instead of an invariant violation.

### Testing
- Existing tests. I probably should add a new one, but again I am hesitant to do so when all the gas numbers are going to change in the near future.

### DIFF
--- a/execution/src/nonces.rs
+++ b/execution/src/nonces.rs
@@ -5,6 +5,7 @@ use {
     move_core_types::{
         account_address::AccountAddress, ident_str, identifier::IdentStr,
         language_storage::ModuleId, resolver::MoveResolver, value::MoveValue,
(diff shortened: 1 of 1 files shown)


ilchu (approved):
I am still curious what causes these gas spikes you've seen, though. But I agree it's up for a later investigation.

PR #319, opened 2025-03-27
Feat(evm-native): Ensure base token balance changes on EVM are reflected on Move

birchmd:
### Description
The goal of this PR was originally just to make deposits via the Optimism Portal contract work properly, but in studying how the deposit mechanism is meant to work it seemed like the right way to achieve this goal was to make a more general change to how base token balances are handled at the Move/EVM boundary.

To that end, this PR adds some an additional hook into the EVM execution to track what transfers are done during the execution. Those same transfers are then also done with the Move representation of our base token. 

### Changes
- Properly handle base token minting in deposit-type transactions (including removing the test-only deposit logic as everything is now unified)
- Trace base token transfers in the EVM and mirror them on Move
- Ensure all deposits done in test code have non-zero `mint` value

### Testing
- Existing tests (including test for deposits)
- Integration test (including a new addition where a deposit is done via the Optimism Portal contract).

### Notes
The changes to `execution/src/tests/gas_cost.rs` are needed to make those tests pass because for some reason changing the deposit flow changed the specific amount of gas that is spend in those tests. I'm not 100% sure why those numbers have changed, but I am not too worried about it since there is a whole workstream related to gas values that we still need to do. If all the gas values are going to change again and be checked for correctness at that time, I won't let it take up my time right now.


evm-ext/src/native_evm_context.rs
```diff
@@ -33,6 +36,7 @@ pub struct HeaderForExecution {
 pub struct NativeEVMContext<'a> {
     pub resolver: &'a dyn MoveResolver<PartialVMError>,
     pub storage_trie: &'a dyn StorageTrieRepository,
+    pub transfers_log: &'a dyn EthTransferLog,
```
bulenttastan, line 39: nit: should it be `transfer_logs` instead of `transfers_log`?

execution/src/deposited.rs
```diff
@@ -95,33 +76,48 @@ pub(super) fn execute_deposited_transaction<
...
+        .and_then(|values| {
+            let evm_outcome = extract_evm_result(values);
+            if !evm_outcome.is_success {
+                return Err(UserError::DepositFailure(evm_outcome.output).into());
+            }
+
+            // If there is a non-zero mint amount then we start by
+            // giving those tokens to the EVM native address.
+            // The tokens will then be distributed to the correct
+            // accounts according to the transferred that happened
```
bulenttastan, line 89: nit: transferred -> transfers

evm-ext/src/events.rs
```diff
@@ -1,15 +1,68 @@
...
+    fn add_tx_origin(&self, address: AccountAddress, amount: U256) {
+        self.origins.borrow_mut().push((address, amount));
+    }
+
+    fn take_origins(&self) -> Vec<(AccountAddress, U256)> {
+        self.origins.take()
+    }
+}
+
+impl EthTransferLog for () {
```
ilchu, line 52: I feel maybe this impl is better under `mod test_doubles` like other similar mock-ish impls we have?

evm-ext/src/native_impl.rs
```diff
@@ -157,6 +175,32 @@ fn evm_transact_inner(
...
+                            amount: *balance,
+                        })
+                    } else {
+                        None
+                    }
+                })
+            });
+        for t in transfers {
+            transfers_log.borrow_mut().push(t);
+        }
```
ilchu, line 200: It seems we can just directly `evm_native_ctx.transfers_log.append_transfers(transfers)` here after collecting them into a vec, or at least the compiler doesn't complain. Am I missing some reason the intermediary refcell is needed?

evm-ext/src/native_impl.rs
```diff
@@ -107,7 +113,19 @@ fn evm_transact_inner(
     // TODO: does it make sense for EVM gas to be 1:1 with MoveVM gas?
     let gas_limit: u64 = context.gas_balance().into();
 
+    // Struct external to the EVM to capture transfer events.
+    // This is used for bookkeeping token balances between Move and EVM.
+    //
+    // It needs to use a `RefCell` for interior mutability because the
+    // EVM handler hooks require the closures to be immutable.
+    // The usage of interior mutability is safe because execution of the single
+    // EVM instance is single threaded.
```
ilchu, line 122: This would probably make a good docstring on the `EthTransfersLogger` struct.

execution/src/canonical.rs
```diff
@@ -201,6 +208,16 @@ pub(super) fn execute_canonical_transaction<
         }
     };
 
+    let vm_outcome = vm_outcome.and_then(|_| {
+        // Ensure any base token balance changes in EVM are reflected in Move too
+        eth_token::replicate_transfers(
```
ilchu, line 213: AFAICT transfers are replicated after getting the VM outcome for both the canonical and deposited tx flow. Any way we could abstract it to a higher level so that it's e.g. called in e.g. `execute_transaction()`? Or does it break the transfer logic?

execution/src/eth_token.rs
```diff
@@ -231,6 +231,48 @@ pub fn transfer_eth<G: GasMeter>(
...
 
+pub fn replicate_transfers<G: GasMeter, L: EthTransferLog>(
+    eth_transfer_logger: &L,
+    session: &mut Session,
+    traversal_context: &mut TraversalContext,
+    gas_meter: &mut G,
+) -> Result<(), moved_shared::error::Error> {
+    // Transfer the transaction value from EVM native account to `origin`.
+    // This step is needed because all EVM transactions start with the caller
+    // transferring tokens to the EVM native account as part of `evm_call`.
```
ilchu, line 242: I was a bit confused about this phrasing and it took me a while to figure out what's happening. Maybe it'd help to specify that this happens as part of the deposited tx execution and we design all mints by default to happen to the EVM native address.

ilchu (commented):
Neat revm stuff I didn't know about. And good peace of mind knowing we're finally doing with transfers what OP itself would be doing.

execution/src/eth_token.rs
```diff
@@ -231,6 +231,48 @@ pub fn transfer_eth<G: GasMeter>(
     Ok(())
 }
 
+pub fn replicate_transfers<G: GasMeter, L: EthTransferLog>(
```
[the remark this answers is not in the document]
birchmd: When a user calls the EVM with an attached value, those tokens are transferred to the EVM account (on Move). The value of the call is recorded in the `EthTransferLog`. The first step of the `replicate_transfers` function is to send the tokens given to the EVM by those calls back to the accounts that made the calls and then proceed with mirroring all the transfers that happened during the EVM execution.

Transferring the tokens from caller to EVM happens to ensure the user has the balance they claim at the time the call is made. Transferring the tokens from EVM back to caller happens to ensure the user has the balance the EVM thinks they had during the execution so that those transfers can happen successfully on the Move side.

In the special case of withdrawal, the user's tokens with ultimately end up at the address of the message passing L2 contract, but inside the MoveVM. If we don't like the idea of tokens accumulating there then we could an additional step which burns the balance at that address.

execution/src/canonical.rs
birchmd: The reason we need to do it here is because it needs to have the session object still in scope. The session is what ultimately creates the MoveVM changes that we return from this function. We could refactor to return the session object up to `execute_transaction`, but I don't think that will make things cleaner because there is different post-processing in the case of canonical vs deposit.

evm-ext/src/native_impl.rs
birchmd: Good catch!

This PR went through a couple iterations and in an earlier version the logger itself was mutable, so it couldn't go here and the intermediary was needed. But that mutability ended up being a problem somewhere else, so I changes the type signature to immutable and now there is no reason for the intermediary. I'll simplify it.

execution/src/canonical.rs
ilchu: I see, thanks for clarifying!

evm-ext/src/events.rs
ilchu: Right, and any other simulation like gas estimation, I believe. Makes sense.

birchmd: Well gas estimation actually does use the real implementation because the extra transfers that happen on the Move side to mirror the EVM cost gas and therefore need to be accounted for in the gas estimation.

evm-ext/src/native_evm_context.rs
birchmd: Changed in https://github.com/MovedNetwork/op-move/pull/319/commits/56a3eb19db003048aa820ecb849ed6d6fceee767

execution/src/deposited.rs
birchmd: Fixed in https://github.com/MovedNetwork/op-move/pull/319/commits/3a899d67ceb68cc39e56b799778256fe6d05d31c

evm-ext/src/native_impl.rs
birchmd: Fixed in https://github.com/MovedNetwork/op-move/pull/319/commits/4b1845a594e661cbbc0db45dc6ce5dd181271738

evm-ext/src/native_impl.rs
birchmd: Moved in https://github.com/MovedNetwork/op-move/pull/319/commits/4b1845a594e661cbbc0db45dc6ce5dd181271738

execution/src/eth_token.rs
birchmd: Additional comment added in https://github.com/MovedNetwork/op-move/pull/319/commits/2ac0015e5aa4e8fc63257354212957ee28c723a6

bulenttastan (approved):
Looks good to me 🚀

PR #320, opened 2025-04-01
refactor: Reduce all application dependencies into one

### LINKED ISSUE #174: Simplify StateActor parameters
Currently `StateActor` takes in many parameters with generic types. We can simplify by merging some existing functionality inside others or merging them into multiple structs.

code-dogg:
Closes #174 

# About
Groups all application dependencies into one object.

# Motivation
* The `StateActor::new` has a lot of parameters. There is a parameter per dependency. Since state actor orchestrates all queries and commands, it is dependent on everything.
* The creation of these dependencies has several variants for different kinds picked by feature flags. It is not standardized by any language tool, but it implicitly relies on the assumption that functions for creating these dependencies are named the same in every variant.

# Problem
* The `StateActor::new` triggers a warning from `clippy` that is supressed.
* `StateActor` is an exclusive owner of all these dependencies. Delegating the responsibility of specifying and holding all application dependencies and making them shareable is a prerequisite for parallel read queries.
* The creation of dependencies relies on an implicit assumption on how the `dependency` modules are structured and how the functions are named.

# Solution
* Create a new object `Application` that holds all application dependencies.
* Define a trait `Dependencies` for creating each of these dependencies.
* Make `StateActor` dependent on `Application`.
* Implement `Dependencies` separately for `InMemory`, `Heed` and `RocksDb` versions.

### DIFF
--- a/api/src/methods/get_payload.rs
+++ b/api/src/methods/get_payload.rs
@@ -45,7 +45,7 @@ mod tests {
         super::*,
         crate::methods::forkchoice_updated,
         alloy::primitives::hex,
-        moved_app::Command,
+        moved_app::{Application, Command, StateActor, TestDependencies},
         moved_blockchain::{
             block::{
                 Block, BlockRepository, Eip1559GasFee, InMemoryBlockQueries,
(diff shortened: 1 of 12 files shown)


birchmd (approved):
Looks good to me! Nice job!

Issue #253, opened 2025-01-30
Research deposit transaction receipt `None` fields


bulenttastan:
Should these receipt fields left as `None`?
Sometimes `None` fields cause big transaction flow issues.

https://github.com/MovedNetwork/op-move/blob/5bff3017103056dbc9c1b2d4a300fc22089fd81e/moved/src/types/transactions.rs#L100

PR #321, opened 2025-04-01
Chore(deps): Latest versions of Rust, revm, alloy

birchmd:
### Description
This PR just updates us to the latest version of Rust (including the new 2024 edition) as well as updates `revm` and `alloy` dependencies to the latest versions. There was a breaking change in the `revm` API which required some refactoring of how the EVM is set up in our native extension, but it still works the same as before. There is a lot of code touched because the formatter made some new changes. However there are no changes to any functionality.

In a follow-up PR I'll work on updating our Aptos dependency to a more recent version.

### Changes
- N/A

### Testing
- Existing tests (including integration test)


evm-ext/src/native_impl.rs
```diff
@@ -121,79 +129,60 @@ fn evm_transact_inner(
...
-        revm::handler::mainnet::output(evm_ctx, result)
-    });
-
-    let outcome = evm.transact().map_err(|e| match e {
-        EVMError::Database(e) => SafeNativeError::InvariantViolation(e),
+    let mut handler = WrappedMainnetHandler {
+        inner: InnerMainnetHandler::default(),
+        evm_native_ctx,
+    };
+    let outcome = handler.run(&mut evm).map_err(|e| match e {
```
bulenttastan, line 184: Nice 😎

execution/src/gas.rs
```diff
@@ -164,6 +164,9 @@ impl L1GasFee for EcotoneGasFee {
             l1_base_fee_scalar: Some(self.base_fee_scalar.saturating_to()),
             l1_blob_base_fee: Some(self.blob_base_fee.saturating_to()),
             l1_blob_base_fee_scalar: Some(self.blob_base_fee_scalar.saturating_to()),
+            // TODO: What are these?
```
bulenttastan, line 167: Currently every TODO has an issue, we can create one for this too.

bulenttastan (approved):

evm-ext/src/native_impl.rs
```diff
@@ -121,79 +129,60 @@ fn evm_transact_inner(
         evm_native_ctx.storage_trie,
         evm_native_ctx.resolver,
     ));
-    // todo: storage trie repository factory?
```
ilchu, line 124: I guess this was pre-EVM storage trie refactor stuff and is now obsoleted?

execution/src/gas.rs
ilchu: We had a very similar one in [transaction.rs](https://github.com/MovedNetwork/op-move/blob/cf972b1d5c69b82d464f9bf615d382d69ffe0270/execution/src/transaction.rs#L102), but I couldn't find an issue for it, actually.

ilchu (commented):
Great work! I took a quick look at the Aptos bump myself, and it seems to be quite a bit more involved with their Loader V2 [changes](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-107.md). Looking forward to seeing that resolved.

execution/src/gas.rs
birchmd: #327 for new TODO.

#253 is the issue corresponding to the `None` fields in transactions.rs.

evm-ext/src/native_impl.rs
birchmd: Yeah, @RomanHodulak would have put that todo there during the EVM trie work, but I didn't see why it was still relevant so it seemed safe to delete. Correct me if I'm wrong Roman!

PR #323, opened 2025-04-02
perf: Remove indirection and heap allocation from callbacks

code-dogg:
# About

Reduces overhead for creating callback functions.

# Motivation

The type for creating callbacks is pretty complex and requires heap allocation for static values.

# Problem

The callback functions, when used as `Box<dyn Fn(&mut self, ...)>` require mutable borrowing of `self`. But calling the function also requires `self` to be borrowed. Therefore, calling it and passing `self` or a field on `self` as argument, means that `self` is borrowed twice, triggering a compile error.

To circumvent this, an indirection is introduced `Box<dyn Fn(...) -> Box<dyn Fn(&mut self)>>` which is a function that creates a function. In this way, the created function is a new value that is not owned by `self` and therefore can accept `self` or fields on `self` as arguments.

# Solution

The callback is reduced to a static reference. In this way, it does not force borrowing of `self`. Without borrowing of `self`, there is no need to have the indirection of creating callback function as its own independently owned object.

### DIFF
--- a/app/src/actor.rs
+++ b/app/src/actor.rs
@@ -47,20 +47,13 @@ use {
 };
 
 /// A function invoked on a completion of new transaction execution batch.
-pub type OnTxBatch<S> =
-    Box<dyn Fn() -> Box<dyn Fn(&mut S) + Send + Sync + 'static> + Send + Sync + 'static>;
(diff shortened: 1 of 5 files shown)


birchmd (approved):
Nice one!

Issue #328, opened 2025-04-03
Use Move Loader V2


birchmd:
AIP: https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-107.md
Implementation PR: https://github.com/aptos-labs/aptos-core/pull/14184 (might be helpful if they update tests which are using public library functions)

Need to:

- [ ] Replace `ResolverBasedModuleBytesStorage` with a proper implementation of the relevant Aptos traits (probably will require a change to our `State` trait)
- [ ] Change `use_v2_loader` to `true` when constructing the MoveVM instances
- [ ] Remove calls to deprecated functions (e.g. `publish_module`)

PR #330, opened 2025-04-03
Chore(deps): Upgrade to Aptos v1.27.2

### LINKED ISSUE #202: Migrate to latest versions of Aptos and MoveVM crates
Move VM is updated with language features with Move 2 and other libraries are also upgraded. Moved should use newer versions to support new features for Testnet.

birchmd:
### Description
Closes #202 

This PR upgrades the Aptos libraries we depend on to v1.27.2. This includes having access to the Move 2 compiler and using the latest Aptos Framework Move modules (which themselves contain Move 2 features).

This biggest change for us is the new `ModuleStorage` API and an instance of this is required on many of the MoveVM functions we use. For now I am constructing a sort of "trivial" instance of a `ModuleStorage` from the `MoveResolver` we already had. This is probably not the best way to do this, so I have created a separate issue for handling it: #328 .

Along with `ModuleStorage`, the functions on MoveVM for publishing modules are now deprecated. For now I have just allowed us to keep using them, but in that same issue I think we should switch to the new way of persisting modules and stop using the deprecated functions.

There were some minor changes to how resources are accessed (`MoveResolver` lost its `get_resource` method) and how values are serialized (`Value` lost its `simple_serialize` method).

All the tests pass, including the integration test!

### Changes
- Create `ResolverBasedModuleBytesStorage` to fulfill `ModuleStorage` API.
- Include `RuntimeEnvironment` in `MovedVM` struct as it is now necessary for creating MoveVM instances.
- Updates to how resources are retrieved from a `MoveResolver`.
- Updates to how (de)serialization of `Value` is done.
- Use Move compiler V2 in our tests.
- Latest Aptos framework package for our genesis process (includes Move V2 contracts).

### Testing
- Existing tests
- Integration test


execution/src/execute.rs
```diff
@@ -119,36 +122,41 @@ pub(super) fn execute_script<G: GasMeter>(
...
-    ];
+    let args: Vec<Vec<u8>> = [
+        (Value::address(*signer), &ADDRESS_LAYOUT),
+        (Value::address(*contract), &ADDRESS_LAYOUT),
+        (Value::u256(value.to_move_u256()), &U256_LAYOUT),
+        (Value::vector_u8(data), &CODE_LAYOUT),
+    ]
+    .into_iter()
+    .map(|(value, layout)| {
+        ValueSerDeContext::new()
```
bulenttastan, line 154: Maybe we should implement a `serialize` helper method. Not necessary, but it could be useful.

bulenttastan (commented):
It looks good to me, I don't see any issues. I'll let @RomanHodulak to also take a look.

bulenttastan (approved):
It looks good to me, I don't see any issues

genesis/src/vm.rs
```diff
@@ -3,25 +3,61 @@ use {
...
-    move_vm_runtime::move_vm::MoveVM,
+    move_vm_runtime::{
+        RuntimeEnvironment, WithRuntimeEnvironment, config::VMConfig, move_vm::MoveVM,
+    },
 };
 
-pub struct MovedVm;
+pub struct MovedVm {
+    env: RuntimeEnvironment,
+}
```
ilchu, line 15: Would extending it to contain the `ResolverBasedBytesStorage` make sense here? This would shave a few lines off the preamble we have anywhere VM-related. And the VM would still be decoupled as it' spawned through the trait. And as another note, the naming has become a bit imprecise. Maybe call it `VMContext` or `VMRuntimeContainer` or the like?

ilchu (approved):

code-dogg (approved):
Nice work! How did you get around the stack overflow issue of the Move compiler?

execution/src/tests/data_type.rs
```diff
@@ -157,8 +157,17 @@ fn test_deeply_nested_type() {
...
+        result.push(module_bytes[33]);
+        result.push(module_bytes[34]);
+
+        // Update next 2 bytes
+        let (x, y) = update_byte(module_bytes[35]);
+        result.push(x);
+        result.push(module_bytes[36] + y);
+
+        // Copy next 17 bytes
+        for b in &module_bytes[37..54] {
```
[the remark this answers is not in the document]
birchmd: I think the Move V2 compiler output a slightly different byte sequence, which is why this needed to change. This test is a little bit brittle in that sense, but it's kind of naturally about the details of the bytecode, so I don't think we can make it more robust.

execution/src/execute.rs
birchmd: Yeah, it might be nice to have a helper trait for this now that it's a little more verbose. But that can be done in a separate PR.

PR #331, opened 2025-04-04
refactor(app): Separate all application scenarios from `StateActor` into `Application`

code-dogg:
# About
Separates application scenarios from `StateActor`

# Motivation
The `StateActor` is responsible for receiving and processing messages. In addition, it is also responsible for implementing whatever scenario the messages are supposed to execute. It should be responsible for only one of these things.

A prerequisite for a follow-up work is to be able to launch scenarios from other ways than going through `StateActor`.

# Problem
The scenarios cannot be executed without going through the `StateActor` pipeline.

# Solution

* Define methods for commands and queries in `Application`
* Move remaining state into `Application`

### DIFF
--- a/api/src/methods/get_payload.rs
+++ b/api/src/methods/get_payload.rs
@@ -116,10 +116,11 @@ mod tests {
 
         let state: StateActor<TestDependencies<_, _, _, _>> = StateActor::new(
             rx,
-            head_hash,
-            0,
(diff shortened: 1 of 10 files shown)


bulenttastan (approved):
I think it looks good.

birchmd (approved):
Looks good to me

PR #332, opened 2025-04-08
Chore(deps): Upgrade tokio to v1.44.2

birchmd:
### Description
I ran `cargo update` again because there is a new version of tokio out that fixes [a bug](https://rustsec.org/advisories/RUSTSEC-2025-0023.html).

### Changes
- `cargo update`

### Testing
- Existing tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -261,7 +261,7 @@ dependencies = [
  "itoa",
  "serde",
  "serde_json",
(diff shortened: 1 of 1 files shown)


code-dogg (approved):
Let's ship it!

Does not seem to be a bug that affects us.

bulenttastan (approved):

Issue #245, opened 2025-01-30
Match fixed bytes type to the L2 contract generation


bulenttastan:
Generate the L2 contracts with the newly created fixed byte type.

https://github.com/MovedNetwork/op-move/blob/fb43657bc4ecc80234f7108e77106d59a55820b6/genesis-builder/src/generate.rs#L181

PR #333, opened 2025-04-09
feat: represent all other sizes of solidity fixed bytes properly

### LINKED ISSUE #237: Support all fixed bytes sizes
Different fixed byte sizes should be constructed between Solidity code and Move. Most commonly used byte sizes are `bytes32` and `bytes4`, but all the sizes from 1 to 32 should be supported.

https://github.com/MovedNetwork/op-move/blob/5bff3017103056dbc9c1b2d4a300fc22089fd81e/evm-ext/src/solidity_abi.rs#L243

ilchu:
### Description

Until now, we were always assuming fixed bytes types coming from/to solidity as having 32 bytes size. While the ABI does indeed pad out to 32 bytes at all times, the size is still important to determine which slice of that word is actually useful. The main challenge was how to make it available at the type level so that we can deduct the size purely from the struct tag during deserialization (i.e. Sol -> Move), as well as how to make it ergonomic for serialization.

Closes #237, #245.

### Changes

- Fixed our treatment of fixed bytes to be more conformant with the ABI spec
- Made `SolidityFixedBytes` generic over a size marker struct
- Added some type-level magic to extract that size info from type layout
- Replaced the generated fixed bytes types in L2 Move contracts with the relevant representation 

### Testing

There was a test touching `abi_encode_params()` and a test doing the same for `abi_decode_params()`, so I just expanded on that.

### Notes

As we introduced MoveV2 in a recent bump, I hoped we could make a nice use case for enums, but the main hurdle turned out to be decorated layouts not being supported yet, which made all nested type info impossible to extract.


evm-ext/src/solidity_abi.rs
```diff
@@ -306,6 +313,29 @@ fn sol_to_value(sv: DynSolValue) -> Value {
...
+
+    match type_name {
+        "B1" => 1,
+        "B2" => 2,
+        "B4" => 4,
+        "B8" => 8,
+        "B16" => 16,
+        "B20" => 20,
+        "B32" => 32,
+        _ => unreachable!("Other sizes are not available and should cause a failure at Move level"),
```
birchmd, line 335: This is reachable code because nothing stops a user from doing something like `evm::as_fixed_bytes<MyStruct>(bytes)` to construct `SolidityFixedBytes` with an unexpected type parameter. Similarly, the `expect` above can be hit by a user via `evm::as_fixed_bytes<u8>(bytes)`.

Solidity is also supposed to support all byte lengths between 1 and 32 inclusive whereas right now we only support powers of 2 and 20. I understand it would be tedious to declare all 32 of the instances using the current approach.

I have an idea to address both of these problems, though the drawback is making the type-level interface quite cumbersome because Move does not support type aliases unfortunately. My idea is to do something similar to the [typenum crate](https://crates.io/crates/typenum/) in Rust where we represent the type-level numbers in binary.

Let's define a marker struct `U5<phantom A, phantom B, phantom C, phantom D, phantom E> {}` which represents a type-level 5-bit number. We also need the marker structs `B0` and `B1` representing the two possible values of a bit. Then we could have, for example, `type Five = U5<B0, B0, B1, B0, B1>`, if only Move supported type aliases.

Now we define the `as_fixed_bytes` function to have the type signature

```
public fun as_fixed_bytes<A, B, C, D, E>(data: vector<u8>): SolidityFixedBytes<U5<A, B, C, D, E>>
```

Notably this type signature constrains the output so that it really will always be a struct (thus making the `struct_tag().expect(...)` above valid.

Now we can convert the `U5` type we get into a number by checking if each bit is equal to `B0` or not and assembling the binary number as we go (some bit fiddling can be done on a `u8` for example to make this efficient). Notably this means we do not need to error out if the user gives a type other than `B0` or `B1` because any type that is not `B0` is mapped to a bit value of 1. Note this means if a user gives wrong types for all the bits then the default will be to max out the size of the fixed bytes, and I think this is the correct default since `bytes32` is such a common Solidity type.

Now there is an awkward off-by one issue here because a 5-bit number can encode the values 0-31 inclusive instead of 1-32 inclusive. So I think we should simply add 1 to the result we get from assembling the given bits in order to get the intended size. This will be confusing for user so we should include some convenience functions:

```
public fun as_fixed_bytes_1(data: vector<u8>): SolidityFixedBytes<U5<B0, B0, B0, B0, B0>>

public fun as_fixed_bytes_2(data: vector<u8>): SolidityFixedBytes<U5<B0, B0, B0, B0, B1>>

public fun as_fixed_bytes_4(data: vector<u8>): SolidityFixedBytes<U5<B0, B0, B0, B1, B1>>

public fun as_fixed_bytes_8(data: vector<u8>): SolidityFixedBytes<U5<B0, B0, B1, B1, B1>>

public fun as_fixed_bytes_16(data: vector<u8>): SolidityFixedBytes<U5<B0, B1, B1, B1, B1>>

public fun as_fixed_bytes_20(data: vector<u8>): SolidityFixedBytes<U5<B1, B0, B0, B1, B1>>

public fun as_fixed_bytes_32(data: vector<u8>): SolidityFixedBytes<U5<B1, B1, B1, B1, B1>>
```

as well as detailed doc comments on the generic `as_fixed_bytes`.

What do you think?

evm-ext/src/solidity_abi.rs
```diff
@@ -288,8 +292,11 @@ fn sol_to_value(sv: DynSolValue) -> Value {
             256 => Value::u256(u.to_move_u256()),
             _ => unreachable!("Only 8, 16, 32, 64, 128 and 256 bit uints are supported by move"),
         },
-        // Move doesn't have fixed-length arrays, the output will be the same regardless of the length
-        DynSolValue::FixedBytes(w, _size) => Value::vector_u8(w.to_vec()),
+        // Packing it back into type-erased `SolidityFixedBytes`
+        // TODO: any way to pass back the size type parameter?
```
birchmd, line 296: No, we can't enforce types here unfortunately. I think it's ok though because we should only reach this point if the ABI-encoded bytes the user gave were successfully deserialized by the alloy library, and so the type parameters the user gave should have to be right.

evm-ext/src/solidity_abi.rs
ilchu: This is both elegant and horrendous at the same time :sweat_smile:.

I think I wrote that function before I realized we have to enforce 32 byte length at all times, and I used to have some helper func in Move that would calculate length based on the marker and bail out if the generic arg was with an unexpected type name.

It pains me a bit that we wouldn't have a singular deserialization entry point though, but I also don't see any other way to make it more convenient for the end user. So I'll try implementing this and expand the test cases to cover the caveats you mentioned.

PR #335, opened 2025-04-10
feat(app): Implement parallel queries

code-dogg:
# About
Parallelizes read-only access to application state using queries.

# Motivation
The separation of queries and commands gives some useful assumptions.

Commands always mutate state, never return value. In general case, it is appropriate for them to be run sequentially in an asynchronous background processing queue.

What about the queries? Queries read state, never mutate it and always return a value. Currently, queries are handled the same way as commands. This is an unutilized optimization opportunity.

# Problem
The queries are enqueued for sequential processing, but could be simply executed with any other queries concurrently as there is no conflict.

The conflict is only with the background command queue, that touches the same state that is being read by the queries.

# Solution
Instead of serializing queries, enqueuing them and waiting for their reply via a one-shot channel, execute them directly on the application state and get their output as a return value.

To prevent conflicts with the command queue, a lock should be added. The lock should permit unique access for the command queue, but otherwise let grant access to any number of queries. An `RwLock` perfectly fits for this use-case.

## Changes
* Removes queries from state message
* Removes all one-shot channels
* Renames StateActor to CommandActor
* Dispatch queries directly without touching the work queue

## Discussion

If you've read the original issue description, then you are aware that it proposed to solve this issue via either forking inside **State Actor**, or introducing a concept called **Read Actor** that would take care of the forking.

Roughly visualized this would work like this:

Read actor

We can see that the network listener forks the incoming requests into individual workers, that are then joined by a common work queue. The Read Actor listens to the queue, just like the network listener listens to incoming requests and forks queries into individual worker threads. The response is then propagated by a one-shot channel.

Upon observing the proposed solution it seemed to me like this is much unneeded work. The work is already forked once, why joining and forking again? Clearly there must be a simpler solution. From this I chased after a newly imagined way.

Direct access

Which eliminated the extra joining and forking, while also getting the response directly. Following this vision is the solution implemented in this PR.


server/src/dependency/heed.rs
```diff
@@ -125,6 +125,7 @@ fn create_db() -> moved_storage_heed::Env {
 
     let env = unsafe {
         EnvOpenOptions::new()
+            .max_readers(20)
```
ilchu, line 128: I suppose it's just a heuristic? Or any specific reason for this being 5x of the RwLock reader max?

server/src/tests/get_proof.rs
```diff
@@ -73,9 +76,14 @@ async fn test_get_proof() -> anyhow::Result<()> {
...
         ]
     });
-    let response: ForkchoiceUpdatedResponseV1 = handle_request(request, &state_channel).await?;
+    let response: ForkchoiceUpdatedResponseV1 =
+        handle_request(request, &state_channel, &app).await?;
     let payload_id = response.payload_id.unwrap();
 
+    state_channel
+        .reserve_many(state_channel.max_capacity())
+        .await?;
```
ilchu, line 85: AFAICT from the docs, this returns a permit iterator that should be called to send data, and as we're reserving a full capacity, all direct calls to the channel from there on should fail?

ilchu (commented):
I'd say it almost doesn't feel like added complexity, at least in terms of how comprehensible code is. Stellar work!

api/src/methods/new_payload.rs
```diff
@@ -29,21 +30,15 @@ async fn inner_execute_v3(
...
-    let (tx, rx) = oneshot::channel();
-    let msg = Query::GetPayloadByBlockHash {
-        block_hash: execution_payload.block_hash,
-        response_channel: tx,
-    }
-    .into();
-    state_channel.send(msg).await.map_err(access_state_error)?;
-    let maybe_response = rx.await.map_err(access_state_error)?;
-
     // TODO: in theory we should start syncing to learn about this block hash.
```
birchmd, line 37: Nit: this comment has moved a few lines further from the error handling it is referencing. Maybe it makes sense to put it inside the `ok_or` now.

server/src/tests/get_proof.rs
birchmd: It's probably worth adding that explanation as a comment in the code since it is a sort of counter-intuitive usage of the reserve_many function.

@ilchu future messages to the channel do not fail because the permit iterator is dropped immediately, so all the capacity is released back.

ilchu: I see, it is indeed a bit counter-intuitive.

PR #338, opened 2025-04-15
perf(blockchain): Remove transactions from blocks

code-dogg:
Reduces space taken by transactions by 50%

### Description

Insert description here

### Changes

- Change 1
- Change 2

### DIFF
--- a/app/src/command.rs
+++ b/app/src/command.rs
@@ -35,54 +35,8 @@ impl<D: Dependencies> Application<D> {
     }
 
     pub fn start_block_build(&mut self, attributes: Payload, id: PayloadId) {
-        let block = self.create_block(attributes);
-        self.block_repository
-            .add(&mut self.storage, block.clone())
-            .unwrap();
-        let block_number = block.block.header.number;
-        let block_hash = block.hash;
-        let base_fee = block.block.header.base_fee_per_gas;
-        self.transaction_repository
-            .extend(
-                &mut self.storage,
-                block
-                    .block
-                    .transactions
-                    .clone()
-                    .into_iter()
-                    .enumerate()
-                    .map(|(transaction_index, inner)| {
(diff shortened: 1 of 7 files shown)


app/src/command.rs
```diff
@@ -147,23 +101,66 @@ impl<D: Dependencies> Application<D> {
...
-        .with_payload_attributes(payload_attributes)
+        .with_payload_attributes(attributes)
         .with_execution_outcome(execution_outcome);
 
-        let hash = self.block_hash.block_hash(&header);
+        let block_hash = self.block_hash.block_hash(&header);
+
+        let block = Block::new(
+            header,
+            op_transactions.iter().map(|v| v.trie_hash()).collect(),
```
birchmd, line 111: I think we can avoid hashing the transactions again here. We already computed all the transaction hashes when collecting which transactions to try to execute.

One idea is to replace this with `receipts.iter().map(|r| r.transaction_hash).collect()`. Another idea is to create a hashes vector from `transactions` before we consume it in `execute_transactions`.

Although I have now noticed that these two approaches are not the same because we simply drop invalid transactions during executions. I.e. there could be some transactions in `op_transactions` that are not in `receipts`. I do not think this is the right behaviour. I think any transaction that is included in a block (i.e. is in its transaction list and included in the transactions root) is supposed to have a receipt. I think this means we should return `op_transactions` from `execute_transactions` instead of constructing it in advance as we do now. Then everything will be consistent and we can use the transaction hashes in `receipts` to fill the block transactions list.

This bug is not related to the issue this PR aims to solve, so it can be fixed in a follow-up PR.

birchmd (approved):
Great find! I like what this PR is doing, though I noticed an unrelated bug while reviewing it. I think this PR can be merged as-is and the bug should be fixed in a follow-up PR.

PR #340, opened 2025-04-17
build(docker): Parse rust toolchain from workspace file

code-dogg:
### Description
* Speeds up the build by downloading the proper toolchain as default
* Removes unused toolchain

### Changes
- Parse toolchain version from `rust-toolchain` file

### Testing
```
docker compose build
```

### DIFF
--- a/docker/Dockerfile.op-move
+++ b/docker/Dockerfile.op-move
@@ -2,6 +2,9 @@ FROM alpine:3.21.2 AS build
 
 WORKDIR /volume
 
+# Copy file with workspace rust toolchain
+COPY rust-toolchain rust-toolchain
+
 RUN \
 # Install build dependencies
     apk upgrade --no-cache \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.op-move
```diff
@@ -2,6 +2,9 @@ FROM alpine:3.21.2 AS build
 
 WORKDIR /volume
 
+# Copy file with workspace rust toolchain
+COPY rust-toolchain rust-toolchain
```
bulenttastan, line 6: The source is already copied below right, is this needed?

code-dogg: This is needed because the sources are copied after the toolchain is installed.

1. Copy `rust-toolchain` file
2. Install toolchain while reading version from `rust-toolchain` file
3. Copy sources
4. Build sources

The order of the operations is adjusted for efficient caching. For example:
* If you change sources, this invalidates step 3 and 4, but step 1 and 2 is skipped.
* If you change toolchain version, it invalidates step 1 and 2, which makes sense because a different toolchain needs to be installed.

bulenttastan: Makes sense, the order matters.

bulenttastan (approved):

PR #341, opened 2025-04-17
feat: Shut the server down if command queue panics

code-dogg:
# About
Maintains the consistency of the server application state.

# Motivation
If the application is running we want some guarantee that it is in a consistent state that is ready to be used.

# Problem
When a panic happens during the processing of a command, it closes the command channel.

However, the server remains running and still accepts requests to RPC endpoints.

In this application state, if a command is sent to the queue, it fails and returns an RPC error. Such case would be valid if that would be a user error. A command should never panic based on a user input, therefore this does not classify as a user error and should not result in a RPC error response.

# Solution
A server is launched while listening to a graceful shutdown listener. That listener listens to a signal that is turned on when the background thread processing the commands exits.

### DIFF
--- a/app/src/queue.rs
+++ b/app/src/queue.rs
@@ -1,0 +1,54 @@
+use {
+    crate::Command,
+    std::pin::Pin,
+    tokio::sync::{broadcast, mpsc},
+};
+
+#[derive(Debug, Clone)]
+pub struct CommandQueue {
+    sender: mpsc::Sender<Command>,
+    killshot: broadcast::Sender<()>,
+}
+
+impl CommandQueue {
+    /// Constructs new [`CommandQueue`] that sends [`Command`]s via `sender`.
+    ///
+    /// In case of a panic of the `sender` channel, a shutdown signal is sent through the
+    /// `killshot`.
+    pub fn new(sender: mpsc::Sender<Command>, killshot: broadcast::Sender<()>) -> Self {
+        Self { sender, killshot }
+    }
+
+    /// Sends a [`Command`] to the background queue for asynchronous processing.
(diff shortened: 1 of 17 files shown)


server/src/lib.rs
```diff
@@ -63,10 +63,7 @@ static JWTSECRET: Lazy<Vec<u8>> = Lazy::new(|| {
     hex::decode(jwt).expect("JWT secret should be a hex string")
 });
 
-pub async fn run() {
-    // TODO: think about channel size bound
-    let (state_channel, rx) = mpsc::channel(1_000);
-
+pub async fn run(max_buffered_commands: u32, max_concurrent_queries: u32) {
```
bulenttastan, line 66: We should keep the `max_buffered_commands` and `max_concurrent_queries` parameters in `GenesisConfig`.

bulenttastan (commented):
I left a comment on moving params to genesis config, the rest of the code looks good. I'll let @birchmd approve if this architectural change with broadcasting shutdown makes sense.

app/src/queue.rs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            self.shutdown();
+        }
+    }
+
+    /// Signals the shutdown.
+    pub fn shutdown(&self) {
+        self.killshot
+            .send(())
+            .expect("Sending shutdown request should succeed");
```
birchmd, line 45: I think it's fine to just have `.ok()` here. The only way this returns an error is if there are no receivers for the message (i.e. nothing that called `shutdown_listener` is still active), but if there are no receivers then there is nothing to shutdown so the error is not a problem.

PR #342, opened 2025-04-19
Fix(tx): Return only values from call without layout

### LINKED ISSUE #322: Convert returned bytes to typed values in SDK
The simulate call or `#[view]` method responds with serialized return values with the type layout. We can create our SDK to simplify the deserialization, along with transaction payload creation and Viem wrappers.

bulenttastan:
### Description
Only return the results of the transaction in bytes without Move value layout. This is the standard way the results are returned even though `session` execution also returns the layout. This response works better when it comes to parsing on the client side. Closes #322 

### Changes
- Return `call_transaction` response bytes.
- Create EVM native outcome layout to be used in integration test.
- Fix integration test to parse the balance and allowance values from the returned bytes.

### Testing
✓ Integration test passes.

### DIFF
--- a/evm-ext/src/lib.rs
+++ b/evm-ext/src/lib.rs
@@ -10,11 +10,15 @@ pub use self::{
 };
 
 use {
+    crate::events::EVM_LOG_LAYOUT,
     move_core_types::{
-        account_address::AccountAddress, ident_str, identifier::IdentStr, value::MoveTypeLayout,
+        account_address::AccountAddress,
+        ident_str,
+        identifier::IdentStr,
+        value::{MoveStructLayout, MoveTypeLayout},
     },
     revm::primitives::Log,
-    std::sync::LazyLock,
+    std::{clone::Clone, sync::LazyLock},
 };
 
 pub mod events;
@@ -34,6 +38,14 @@ pub const EVM_NATIVE_MODULE: &IdentStr = ident_str!("evm");
 /// Layout for EVM byte code. It is simply a byte vector because we store the raw bytes directly.
 pub static CODE_LAYOUT: LazyLock<MoveTypeLayout> =
     LazyLock::new(|| MoveTypeLayout::Vector(Box::new(MoveTypeLayout::U8)));
+
+pub static EVM_NATIVE_OUTCOME_LAYOUT: LazyLock<MoveTypeLayout> = LazyLock::new(|| {
+    MoveTypeLayout::Struct(MoveStructLayout::Runtime(vec![
+        MoveTypeLayout::Bool,
+        CODE_LAYOUT.clone(),
+        MoveTypeLayout::Vector(Box::new(EVM_LOG_LAYOUT.clone())),
+    ]))
+});
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct EvmNativeOutcome {
--- a/evm-ext/src/lib.rs
+++ b/evm-ext/src/lib.rs
@@ -18,7 +18,7 @@ use {
         value::{MoveStructLayout, MoveTypeLayout},
     },
     revm::primitives::Log,
-    std::{clone::Clone, sync::LazyLock},
+    std::sync::LazyLock,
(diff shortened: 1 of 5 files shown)


execution/src/simulate.rs
```diff
@@ -115,7 +115,13 @@ pub fn call_transaction(
                 verify_input.traversal_context,
                 &code_storage,
             )?;
-            Ok(bcs::to_bytes(&outcome.return_values)?)
+            Ok(bcs::to_bytes(
+                &outcome
+                    .return_values
+                    .into_iter()
+                    .map(|(bytes, _ty)| bytes)
```
ilchu, line 122: Feels it could be useful to state the rationale here, just like in the PR description.

server/src/tests/integration/erc20.rs
```diff
@@ -239,9 +239,14 @@ pub async fn l2_erc20_balance_of(token: Address, account: Address, rpc_url: &str
...
 
+    // Extract the first result and combine with EVM result layout for deserialization
+    let output: Vec<Vec<u8>> = bcs::from_bytes(&eth_call_result)?;
     let return_values = SerializedReturnValues {
         mutable_reference_outputs: Vec::new(),
-        return_values: bcs::from_bytes(&eth_call_result)?,
+        return_values: Vec::from([(
+            output.first().unwrap().to_owned(),
+            EVM_NATIVE_OUTCOME_LAYOUT.clone(),
+        )]),
```
ilchu, line 249: My only concern is whether we would have to keep supplying the layout still, despite not having it on the client side for some internal things like this. And whether such spots would grow when we add EVM NFT support etc.

ilchu (approved):

server/src/tests/integration/erc20.rs
bulenttastan: I checked this and we only do this in the integration test. Other tests perform the `session.execute_function_bypass_visibility()` calls directly without going through the `call` API. So I think this won't be an issue in general.

execution/src/simulate.rs
bulenttastan: Good idea, added in [3ad3a76](https://github.com/MovedNetwork/op-move/pull/342/commits/3ad3a76e161100b4996b3dc323fbd00b25563b48)

execution/src/simulate.rs
```diff
@@ -115,7 +115,15 @@ pub fn call_transaction(
...
             )?;
-            Ok(bcs::to_bytes(&outcome.return_values)?)
+            // Only return the results of the transaction in bytes without the Move value layout.
+            // Sending just the bytes works better when it comes to parsing on the client side.
+            Ok(bcs::to_bytes(
+                &outcome
+                    .return_values
+                    .into_iter()
+                    .map(|(bytes, _ty)| bytes)
+                    .collect::<Vec<_>>(),
```
birchmd, line 125: I wonder if we could avoid the `collect` here. Can you pass an iterator to `to_bytes` directly?

bulenttastan: It doesn't accept I don't think: 
```
Trait `Serialize` is not implemented for Map<IntoIter
```

evm-ext/src/lib.rs
```diff
@@ -8,11 +8,15 @@ pub use self::{
...
     move_core_types::{
-        account_address::AccountAddress, ident_str, identifier::IdentStr, value::MoveTypeLayout,
+        account_address::AccountAddress,
+        ident_str,
+        identifier::IdentStr,
+        value::{MoveStructLayout, MoveTypeLayout},
     },
     revm::primitives::Log,
-    std::sync::LazyLock,
+    std::{clone::Clone, sync::LazyLock},
```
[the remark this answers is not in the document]
bulenttastan: Done in [63862c7](https://github.com/MovedNetwork/op-move/commit/63862c7de4db02e621bcf4b48307310b9c293499)

birchmd (approved):

Issue #42, opened 2024-08-08
Optimize async code performance


birchmd:
- Benchmarks
- Load testing

Code reference: https://github.com/MovedNetwork/op-move/blob/79df431c0deb8cfe59303d13adff50cbff10ec62/moved/src/main.rs#L61

Note: the question of channel size is related to async performance because tokio runtime prioritizes tasks based on backpressure from channel queues.

PR #345, opened 2025-04-22
test: Add benchmark for command queue size

code-dogg:
### Description
Adds benchmark of performance characteristics of processing a large batch of commands vs a differently sized command queue. 

Part of #42 

### Changes
- Add benchmark for command queue size testing `max_buffered_commands` parameter

### Testing
```
cargo bench
```

#### Violin plot
violin

This chart shows the relationship between function/parameter and iteration time. The thickness of the shaded region indicates the probability that a measurement of the given function/parameter would take a particular length of time.

#### Full report
[Build 1000 blocks with queue size Summary.pdf](https://github.com/user-attachments/files/19868619/Build.1000.blocks.with.queue.size.pdf)

### Discussion
The queue size affects the total time to process all commands significantly.

From the violin plot, we can observe that the performance is worse when the amount of commands outsize the queue size. This can be explained by the fact that every time there is no room in the queue, the future is pending until a slot frees up. This increases the amount of total work needed as this gives the async runtime more work in shuffling around the ongoing tasks.

The performance reaches an optimum when the queue size is greater than or equal to the amount of commands. There is no significant difference past the threshold of 3960.

The question is how accurate is the benchmark. The input comes from a real life usage except for the fact that there is no organic traffic and it's all op-node system transactions. Furthermore, the timing is not realistic, because the input commands are sent immediately as quick as possible. To make this more real, it would need to add a 1 second sleep before every `StartBlockBuild` command. However, that would increase the time it takes to measure the performance by 1000 seconds per queue size, with 10 different sizes that gives us roughly **2 hours 45 minutes**.

Another question is that sending the commands should perhaps not be included in the measured code as that does not influence the performance of processing the commands, rather increases the response time in the RPC handler that invoked it.

### Conclusion
The results suggest that the queue size does not affect the performance as long as it is large enough to contain all the commands.


server/benches/perf/queue/input.rs
birchmd, line 1: I worry a bit about needing to maintain a file this large. I wonder if there is a way it could be procedurally generated as-needed (i.e. only when running the benchmark) rather than directly taken from op-node data (though using that as a starting point makes sense).

One thing I notice is the `UpdateHead` commands are always repeated. I wonder if we could somehow change our logic to remove the redundant one, but that is unrelated to this PR.

More to the point, there is a clear pattern of commands, so perhaps we only need to keep the hex-encoded op-node transactions and maybe the rest of this file could be generated from that one (static) resource.

birchmd (commented):
Thanks for providing the detailed report! `criterion` produces some nice graphs!

> queue size does not affect the performance as long as it is large enough to contain all the commands

This conclusion makes complete sense. The question then becomes: what is the average rate at which we receive commands (let's call it R) vs the rate at which we process them (let's call it P). If R < P then the equilibrium state is that the queue is empty and the capacity is not very important. If R > P then the queue will be perpetually full, regardless of how large we allow it to be so the limitation becomes how much memory we are willing to spend on queued commands. R = P is an edge case where the details of how the rates fluctuate will matter.

Does the violin plot say that even with a queue size of 1 we can produce 1000 blocks in 15-20 ms? And with any reasonable queue size we can produce 1000 blocks in less than 5 ms? If so then I think it is clear that we will be in the case where R < P and the queue size is not very important.

Of course this is an idealized scenario where there are no queries taking the state lock from time to time. But even still, it seems like we should always be able to empty the queue pretty quickly.

If we are worried about memory bounds; the size of the `Command` type is currently 816 bytes. For the sake of estimating we can round this up to 1 kB. If we make the queue capacity 5000 (larger than the number of commands needed to produce 1000 blocks) then we are only allocating 5 MB of RAM to this purpose. If we're expecting to run on a machine with multiple GBs of RAM then I think we can reasonably allocate 5 MB to this purpose.

server/benches/perf/queue/input.rs
code-dogg: Generating this file would be superior as it would remove the maintenance burden and I'm confident that it is possible.

Ad `UpdateHead`: That's a good observation, something worth looking into.

code-dogg:
Thanks for the thoughtful review!

> Does the violin plot say that even with a queue size of 1 we can produce 1000 blocks in 15-20 ms? And with any reasonable queue size we can produce 1000 blocks in less than 5 ms? If so then I think it is clear that we will be in the case where R < P and the queue size is not very important.

It is rather saying that it takes 5 ms / 15-20 ms on average per command. There are 3960 commands of which 1000 are `StartBlockBuild`. Based on this assumption, it should say ~20 seconds / 60-80 seconds for 1000 blocks.

Even then I find it surprisingly fast.

Based on the payload timestamps it looks like R = 4 commands per second.

If average per command processing rate is:
* 5 ms => P = 200 commands per second
* 15 ms => P = 66.666 commands per second
* 20 ms => P = 50 commands per second

Which gives us R < P in all cases.

server/benches/perf/queue/input.rs
code-dogg: Solved.

I looked into `UpdateHead` and the outcome is #348

code-dogg:
@birchmd Your interpretation of the benchmark was correct, mine was not. So we are able to build 1000 blocks in a few milliseconds.

After #348 the rate of received commands is about 1 per second

Here are the new benchmark results [Build 1000 blocks with queue size.pdf](https://github.com/user-attachments/files/19929188/Build.1000.blocks.with.queue.size.pdf)

violin

We're able to build 1000 blocks in about 2.5 - 3ms so the processing rate is between 300000 and 400000 blocks per second.

birchmd (approved):
Great job!

Issue #346, opened 2025-04-23
Charge gas for module deployment


ilchu:
With the introduction of Aptos loader v2, gas is not automatically charged anymore for module deployment transactions. We need to think of a process for that.

PR #347, opened 2025-04-23
feat: enable loader v2

### LINKED ISSUE #328: Use Move Loader V2

ilchu:
### Description

This PR removes use of deprecated aptos functionality and fully transitions to their loader V2 implementation.

Closes #328.

### Changes

- Enables the loader v2 flag
- Uses the same staging module storage for publishing as Aptos does
- Refactored on-chain module deployment

### Testing

All the old tests pass.

### Notes

As the new process doesn't involve a session, gassing module deployment in runtime is not a given anymore. Created #346 to track that.


genesis/src/config.rs
```diff
@@ -76,7 +76,7 @@ impl Default for GenesisConfig {
         Self {
             chain_id: CHAIN_ID,
             initial_state_root: B256::from(hex!(
-                "0xb97674627a3557d273abcd3d80fd01a3bbc0d52c5ebd7e6897f06f5ac97a4e27"
+                "a5d46b471b3d197a9e2cc7996fa9ce02ab7d9f3abb2910ba36618db91d78c67e"
```
bulenttastan, line 79: nit: I think we should keep the original hex format as starting with `0x`

birchmd (approved):
Looks good to me, nice work!

Issue #63, opened 2024-08-08
Implement ForkchoiceUpdate according to the Ethereum specification


birchmd:
The Ethereum spec includes validation conditions on the input to the RPC method which we currently do not have implemented.

Code reference: https://github.com/MovedNetwork/op-move/blob/79df431c0deb8cfe59303d13adff50cbff10ec62/moved/src/methods/forkchoice_updated.rs#L66

Issue #303, opened 2025-03-06
Explore multi-node setup


birchmd:
Currently our tests only run with a single op-node. Need to explore how things work when there is a network of them. We will probably need multiple nodes behind a load balancer to scale our APIs to real amounts of traffic for example.

Issue #225, opened 2025-01-30
Generate state root hash for genesis block internally


bulenttastan:
Currently we run `op-geth` to get the genesis state root hash. We should find a way to create this hash internally so we can completely stop relying on `op-geth`

https://github.com/MovedNetwork/op-move/blob/5065d42637a91b01a04a2c4608e547a58e71d266/server/src/lib.rs#L289

PR #348, opened 2025-04-25
perf(app): Remove `UpdateHead` command

code-dogg:
### Description

While observing the commands received in the app as part of making benchmarks I noticed that the `UpdateHead` command is called repeatedly in unbroken sequences with the same hash.

The command is invoked when calling `engine_forkchoiceUpdated` by `op-node`. This endpoint is called repeatedly without payload as a part of canonical chain validation process. Every time this endpoint is invoked it produces `UpdateHead` command.

This behavior is incorrect, but that is expected and recorded in the issue #63. Perhaps this way was to instead of validation assume that the op-node state is valid and make sure op-move is at the same state.

Effectively this only does one thing which is sets the head hash to the last block. It is critical that this happens. If the head pointer is not correctly updated, building block panics.

Building the block should update the head and validating forkchoice state should not update the state, hence this change. This gives us the following results:

* Reduces the amount of commands received by ~75%
* Reveals bugs in tests where the head is not updated and multiple blocks are built with genesis as parent
* Reveals another bug that causes division by zero when payload received had `gas_limit` set to zero

### Changes
- Update the head block hash pointer only when building new block
- Fix error where having `gas_limit` causes division by zero

### Testing
```
cargo test
```

### DIFF
--- a/api/src/methods/mod.rs
+++ b/api/src/methods/mod.rs
@@ -126,8 +126,11 @@ pub mod tests {
 
         let mut encoded = Vec::new();
         tx.encode(&mut encoded);
-        let mut payload_attributes = Payload::default();
-        payload_attributes.transactions.push(encoded.into());
+        let payload_attributes = Payload {
+            gas_limit: U64::MAX,
+            transactions: vec![encoded.into()],
+            ..Default::default()
+        };
 
         let msg = Command::StartBlockBuild {
             payload_attributes,
@@ -156,8 +159,11 @@ pub mod tests {
 
         let mut encoded = Vec::new();
         tx.encode(&mut encoded);
-        let mut payload_attributes = Payload::default();
-        payload_attributes.transactions.push(encoded.into());
+        let payload_attributes = Payload {
+            gas_limit: U64::MAX,
+            transactions: vec![encoded.into()],
+            ..Default::default()
+        };
 
         let msg = Command::StartBlockBuild {
             payload_attributes,
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -6,7 +6,7 @@ use {
     moved_api::method_name::MethodName,
     moved_app::{Application, Command, CommandQueue, Dependencies},
     moved_blockchain::{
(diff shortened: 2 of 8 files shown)


api/src/methods/mod.rs
```diff
@@ -126,8 +126,11 @@ pub mod tests {
 
         let mut encoded = Vec::new();
         tx.encode(&mut encoded);
-        let mut payload_attributes = Payload::default();
-        payload_attributes.transactions.push(encoded.into());
+        let payload_attributes = Payload {
+            gas_limit: U64::MAX,
```
bulenttastan, line 130: @birchmd is this right?

bulenttastan (approved):
Looks about right!

server/src/lib.rs
```diff
@@ -163,11 +163,8 @@ pub fn initialize_app(genesis_config: GenesisConfig) -> Application<dependency::
     );
 
     let genesis_block = create_genesis_block(&app.block_hash, &genesis_config);
-    let head = genesis_block.hash;
-    app.block_repository
-        .add(&mut app.storage, genesis_block)
-        .expect("Database should be ready");
-    app.update_head(head);
+
+    app.genesis_update(genesis_block);
```
ilchu, line 167: A bit of a side note, but is it right to call it `genesis_update`? AFAICT we're doing the same operations in every block build, i.e. updating head and adding to the block repo. Which means we could call it in `start_block_build` epilogue as well.

ilchu (approved):

birchmd (approved):
Looks good to me.

This interacts with #303 in that if we did have multiple nodes running then we might need to sync the head of the chain from another node. But we can cross that bridge when we start thinking seriously about designing a syncing protocol.

server/src/lib.rs
code-dogg: I see your point. This command is designated for updating genesis from op-geth, but it really just adds a block and updates the head pointer.

I would say that it is better to keep it named this way as it is aligned with its purpose, regardless of what it is doing and also this command should be removed with #225

PR #349, opened 2025-04-28
Feat(evm): allow all users to deploy EVM contracts

birchmd:
### Description
The main goal of this contract is to allow any user to deploy a contract to the EVM. This required several changes as described below.

### Changes
- Rename `SciptOrModule` to `ScriptOrDeployment` and add a variant for deploying EVM contracts.
- New system-only function in EVM Move contract for deploying user contracts
- Handle EVM account self-destruct according to EIP-6780
- `BLOCKHASH` EVM opcode implementation
- Refactor `test_get_proof` to make it easier to reuse

### Testing
- Existing tests
- New `test_evm_contracts` pseudo-integration test.


code-dogg (approved):
Pretty cool, a lot of hard work went into this.

So what happens when you deploy EVM contract? It gets allocated an EVM address and you can pass calls to it from Move via `entry_evm_call`?

birchmd:
Yes, exactly.u

PR #353, opened 2025-04-30
refactor: use op_alloy tx envelopes everywhere

### LINKED ISSUE #85: Use `op-alloy` crate for transaction types
`op-alloy` crate adds deposit transaction type on top of the existing types in `alloy`.

So for instance, instead of creating our own `ExtendTxEnvelope` we can use `OpTransactionEnvelope`. This mainly affects the following files:
* `types/transactions.rs`
* `move-execution.rs`
* `state_actor.rs`

### LINKED ISSUE #253: Research deposit transaction receipt `None` fields

ilchu:
### Description

This PR refactors the envelopes for transactions to rely more on the `op-alloy` crate. It also fixes the nonces not being set properly in the receipts. While the latter could be a separate PR, though a microscopic one, I started by investigating if it would be solved automatically with alloy types, but in the end it didn't. At that point separating the two changes seemed too much of a chore.

Closes #85, closes #253.

### Changes

- The deposit nonces are set properly from the underlying L1 tx as they should be per Optimism spec since Canyon update
- We now reuse the `OpTxEnvelope` instead of having our own custom envelopes

### Testing

- ✓ Old tests passing


api/src/methods/mod.rs
```diff
@@ -111,16 +109,16 @@ pub mod tests {
...
+            to: TxKind::Call(to),
             value: amount,
             source_hash: FixedBytes::default(),
             from: to,
-            mint: amount,
-            gas: U64::from(u64::MAX),
-            is_system_tx: false,
-            data: Vec::new().into(),
-        });
+            mint: Some(amount.try_into().unwrap()),
```
bulenttastan, line 117: I think we can set the amount above as, instead of casting it here:
```rust
let amount = Some(parse_ether("1").unwrap().into());
```

execution/src/transaction.rs
```diff
@@ -18,187 +17,65 @@ use {
...
+            Self::Deposit(dep) => {
                 OpReceiptEnvelope::Deposit(OpDepositReceiptWithBloom {
                     receipt: OpDepositReceipt {
                         inner: receipt,
-                        // TODO: what are these fields supposed to be?
-                        deposit_nonce: None,
-                        deposit_receipt_version: None,
+                        // Per OP stack spec <https://specs.optimism.io/protocol/deposits.html#execution>
+                        deposit_nonce: Some(dep.nonce()),
+                        deposit_receipt_version: Some(1),
```
bulenttastan, line 53: nit: we can make the version number a constant: `const DEPOSIT_RECEIPT_VERSION: u64 = 1;`

birchmd (approved):
Nice job!

PR #354, opened 2025-05-01
Feat(evm): charge gas for solidity ABI native

### LINKED ISSUE #233: Charge gas for Solidity ABI operations
Solidity ABI encoding and decoding are new operations created. These operations should consume appropriate gas.

https://github.com/MovedNetwork/op-move/blob/5bff3017103056dbc9c1b2d4a300fc22089fd81e/evm-ext/src/solidity_abi.rs#L72

https://github.com/MovedNetwork/op-move/blob/5bff3017103056dbc9c1b2d4a300fc22089fd81e/evm-ext/src/solidity_abi.rs#L97

birchmd:
### Description
Charges gas for using the solidity ABI natives.

Closes #233 

### Changes
- Refactor ABI native implementation to make it more testable
- New test for gas costs (including check that values match when encoded then decoded)

### Testing
- New test `solidity_abi::tests::test_gas_costs`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ aptos-table-natives = { git = "https://github.com/aptos-labs/aptos-core", tag = 
 aptos-types = { git = "https://github.com/aptos-labs/aptos-core", tag = "aptos-node-v1.27.2" }
 aptos-vm = { git = "https://github.com/aptos-labs/aptos-core", tag = "aptos-node-v1.27.2" }
 aptos-vm-types = { git = "https://github.com/aptos-labs/aptos-core", tag = "aptos-node-v1.27.2" }
+arbitrary = "1"
 bcs = { git = "https://github.com/aptos-labs/bcs" }
 better_any = "0.1.1"
 bytes = "1.6"
@@ -80,6 +81,7 @@ moved-trie = { path = "trie" }
 once_cell = "1.19"
 op-alloy = { version = "0.12", features = ["full", "std", "k256", "serde"] }
 openssl = "0.10"
+rand = "0.8"
 regex = "1.11.1"
 revm = { version = "21", features = ["optional_balance_check"] }
 rocksdb = { version = "0.23" }
--- a/evm-ext/src/solidity_abi.rs
+++ b/evm-ext/src/solidity_abi.rs
@@ -2,10 +2,18 @@ use {
     super::{EVM_NATIVE_ADDRESS, EVM_NATIVE_MODULE},
     crate::native_evm_context::FRAMEWORK_ADDRESS,
     alloy::dyn_abi::{DynSolType, DynSolValue, Error},
+    aptos_gas_algebra::{GasExpression, InternalGasUnit},
+    aptos_gas_schedule::{
+        NativeGasParameters,
+        gas_params::natives::aptos_framework::{TYPE_INFO_TYPE_OF_BASE, UTIL_FROM_BYTES_PER_BYTE},
+    },
     aptos_native_interface::{
         SafeNativeContext, SafeNativeError, SafeNativeResult, safely_pop_arg, safely_pop_type_arg,
     },
+    aptos_types::vm_status::StatusCode,
+    move_binary_format::errors::PartialVMError,
     move_core_types::{
+        gas_algebra::NumBytes,
         ident_str,
         language_storage::{StructTag, TypeTag},
         value::{MoveStructLayout, MoveTypeLayout, MoveValue},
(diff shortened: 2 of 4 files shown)


evm-ext/src/solidity_abi.rs
```diff
@@ -299,18 +369,50 @@ fn sol_to_value(sv: DynSolValue) -> Value {
         DynSolValue::FixedBytes(w, _size) => {
             Value::struct_(Struct::pack(vec![Value::vector_u8(w.to_vec())]))
         }
-        DynSolValue::Address(a) => Value::address(a.to_move_address()),
+        DynSolValue::Address(a) => match layout {
+            MoveTypeLayout::Address => Value::address(a.to_move_address()),
+            MoveTypeLayout::Signer => Value::master_signer(a.to_move_address()),
```
bulenttastan, line 374: Nice catch, address could be a signer.

birchmd: Actually, thinking about this again, it shouldn't be allowed to create a signer because then you could use this to forge a signature of any address. Since the main purpose of this API is to parse data coming back from EVM calls, there should never be a signer object to decode. I'll replace this with an error instead to make sure it cannot happen.

Cargo.toml
```diff
@@ -80,6 +81,7 @@ moved-trie = { path = "trie" }
 once_cell = "1.19"
 op-alloy = { version = "0.12", features = ["full", "std", "k256", "serde"] }
 openssl = "0.10"
+rand = "0.8"
```
bulenttastan, line 84: Just curious, there is no `[workspace.dev-dependencies]` right?

evm-ext/src/solidity_abi.rs
```diff
@@ -69,11 +77,23 @@ pub fn abi_encode_params(
     let prefix = args.pop_back().unwrap().value_as::<Vec<u8>>()?;
     let ty_arg = safely_pop_type_arg!(ty_args);
 
-    // TODO: need to figure out how much gas to charge for these operations.
+    // Charge for the lookup of the type (twice because we need to get annotated and not).
```
bulenttastan, line 80: I'm sure this is correct, but I don't fully understand the explanation here. Is it twice for annotated and unannotated?

Cargo.toml
birchmd: Correct, there is not.

evm-ext/src/solidity_abi.rs
birchmd: Yeah we need to look up both layouts so we need to charge once for each. The users must pay for all resource expenditures. If we can find a more efficient implementation of this in the future then we could lower its gas cost.

evm-ext/src/solidity_abi.rs
```diff
@@ -94,34 +114,84 @@ pub fn abi_decode_params(
     let value = safely_pop_arg!(args, Vec<u8>);
     let ty_arg = safely_pop_type_arg!(ty_args);
 
-    // TODO: need to figure out how much gas to charge for these operations.
+    // Charge for the lookup of the type.
+    context.charge(TYPE_INFO_TYPE_OF_BASE)?;
+    // Charge gas for decoding. It is important to charge _before_ we do the work.
+    context.charge(abi_decode_gas_cost(&value))?;
```
code-dogg, line 120: Having a function that charges gas and performs the operation would make it easier to ensure the same cost is always charged.

code-dogg (approved):
Fast delivery :muscle:

evm-ext/src/solidity_abi.rs
birchmd: I did try something along those lines, but it was more complex to write a test for. The issue is that `SafeNativeContext` is not a publicly constructable type (all the ways to make it are only visible in the Aptos crate), so then I would need to create some trait that abstracts over the gas charging and use a different implementation type in testing.

I didn't think defining a new trait was worthwhile when all functions that do not charge gas are private to this module. The only public functions are `abi_encode_params` and `abi_encode_params` which do charge gas via `SafeNativeContext`.

evm-ext/src/solidity_abi.rs
birchmd: Done in https://github.com/MovedNetwork/op-move/pull/354/commits/5f2f40a980db0f129317b5386fee15d411713df7

evm-ext/src/solidity_abi.rs
ilchu: I think we don't have an issue for investigating this annotated layout not being supported, right? And I feel the aptos version bump didn't change things.

evm-ext/src/solidity_abi.rs
```diff
@@ -355,3 +464,243 @@ fn force_to_u8(mv: MoveValue) -> u8 {
...
+        alloy::primitives::Address,
+        aptos_gas_schedule::{InitialGasSchedule, LATEST_GAS_FEATURE_VERSION},
+        arbitrary::Unstructured,
+        move_core_types::value::{MoveFieldLayout, MoveStruct},
+        rand::{Rng, RngCore, rngs::ThreadRng, seq::SliceRandom},
+        std::time::{Duration, Instant},
+    };
+
+    #[test]
+    fn test_gas_costs() {
```
ilchu, line 481: Neat and fuzzy ✨ But I guess no other tests that would benefit from this approach?

ilchu (approved):

evm-ext/src/solidity_abi.rs
```diff
@@ -283,36 +353,75 @@ fn layout_to_sol_type(layout: &MoveTypeLayout) -> Result<DynSolType, Error> {
...
+        (DynSolValue::Bytes(b), MoveTypeLayout::Vector(inner_layout))
+            if matches!(inner_layout.as_ref(), MoveTypeLayout::U8) =>
+        {
+            Value::vector_u8(b)
+        }
+        (DynSolValue::String(s), MoveTypeLayout::Struct(_)) => {
             Value::struct_(Struct::pack(vec![Value::vector_u8(s.into_bytes())]))
         }
-        DynSolValue::Array(a) => Value::vector_for_testing_only(
-            // TODO: Make sure all the vector elements are of the same type
```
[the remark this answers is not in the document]
birchmd: Yes, having access to the layout is why the types are now checked in recursive calls. Thanks for pointing out it closes another issue! 🚀

evm-ext/src/solidity_abi.rs
birchmd: There is no issue for it. Honestly I don't think it's a problem; the gas costs shouldn't be too bad even with the double type info lookup. Nothing about this has changed with the recent Aptos version bump, correct.

evm-ext/src/solidity_abi.rs
birchmd: Potentially other tests would benefit! It might be nice to review and check. The difference here is I think this test could not have been written adequately without this approach. Getting gas costs right is important because once they are set we can only lower them as raising them might break existing code and we do not want to do that.

PR #355, opened 2025-05-02
feat(server): Recover from persisted state

code-dogg:
Following-up on #350 the node can simply start-up and recover from persisted state exactly where it stopped. Strictly speaking, the only thing that can be lost is the transactions mem-pool if it's not empty.

However, this still does not grant reliable re-deployment of upgraded versions of the op-move container. The reason is that the op stack may not fully recover past the brief unavailability of the execution client.

The next step is to investigate ways to grant that the op stack survives op-move restart.
1. Maybe there are ways to improve graceful shutdown?
2. Implement zero-downtime deployment configuration

Specifically the option 2 should be the best because if the node restores from persisted state and there is no unavailability, then it's the same as if the node is running continuously. This also improves the UX as the RPC does not experience any downtime.

### DIFF
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -6,7 +6,7 @@ use {
     moved_api::method_name::MethodName,
     moved_app::{Application, Command, CommandQueue, Dependencies},
     moved_blockchain::{
-        block::{Block, BlockHash, ExtendedBlock, Header},
+        block::{Block, BlockHash, BlockQueries, ExtendedBlock, Header},
         payload::{NewPayloadId, StatePayloadId},
     },
     moved_genesis::config::GenesisConfig,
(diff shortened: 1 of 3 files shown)


birchmd (approved):
Maybe in the future we'll want a command line flag in addition to environment variable, but this looks like a good starting point.

server/src/lib.rs
```diff
@@ -138,32 +138,34 @@ pub async fn run(max_buffered_commands: u32, max_concurrent_queries: u32) {
...
-    };
-    moved_genesis::apply(
-        genesis_changes,
-        table_changes,
-        evm_storage_changes,
-        &genesis_config,
-        &mut app.state,
-        &mut app.evm_storage,
-    );
+    if app.block_queries.latest(&app.storage).unwrap().is_none() {
```
ilchu, line 141: Feels good how it's pretty much a line change because everything was in place already.

ilchu (approved):
By having zero downtime do you suggest e.g. winding up replicas of op-move or some other strategy like that?

code-dogg:
@ilchu Yes, some kind of way where 
1. new version of op-move is deployed to a new container
2. the traffic from op stack gets redirected to the new container
3. old container is removed

For example, using an orchestration engine with load balancer where you can scale up and down the number of replicas.

The main problem is that the application container is bound to an exclusive port number. If you deploy a new container, while it can use the same port inside the container, it cannot expose it to the host. Therefore, there either needs to be a proxy or a load balancer service that allocates the host port, or some kind of remapping of host ports.

Issue #327, opened 2025-04-03
Set values for operator fee


birchmd:
New fields have been introduced into the `L1BlockInfo` struct. They are optional so they are currently set to `None`, but we know that is not necessarily the correct value. This task is to figure out what the right values are for `operator_fee_scalar` and `operator_fee_constant`. [This link](https://gov.optimism.io/t/upgrade-proposal-15-isthmus-hard-fork/9804) may have useful information in making a decision. These fields were introduced into our code in #321 .

PR #356, opened 2025-05-02
chore(gas): reword operator fee todo for clarity

ilchu:
### Description

After research into the OP governance docs regarding operator fees, I found out that these fields are indeed set at L1 level in the SystemConfig contract, but their storage slots and values are still 0, though it might change if some consensus baseline for them is agreed upon in the future. So #327 still needs to stay open, but I demoted it to P2 as we're not doing anything wrong yet.

### Changes

- Clearer wording for the TODO


bulenttastan (approved):
Looks good. Unit tests failed for some reason, should look into that.

PR #358, opened 2025-05-13
Feat(evm): Add evm_view function to Move EVM interface

### LINKED ISSUE #177: Execute EVM call without a signer for no state change methods
When calling L2 contracts through the Move interface the caller always has to send in the `&signer` on the [contract generation](https://github.com/MovedNetwork/op-move/blob/main/genesis-builder/l2_move_template.hbs#L30).

The interface should not need the `signer` for `getter` methods, but the `Evm.move` doesn't have the right method to skip the signer. We should have a way to call these getter methods without the `signer` param.

birchmd:
### Description
Closes #177 

Makes it easier to call view methods of EVM contracts.

### Changes
- New `evm_view` native function
- Updates to ERC-20 and L2 contract Move interfaces

### Testing
- Existing test (updated to use the new `evm_view` where appropriate).

### DIFF
--- a/genesis-builder/l2_move_template.hbs
+++ b/genesis-builder/l2_move_template.hbs
@@ -1,10 +1,7 @@
 module {{name}}::{{snake name}} {
     {{#if has_fungible_asset}}
-    {{#if has_eth_token_metadata}}
-    use aptos_framework::fungible_asset_u256::{FungibleAsset, zero};
-    {{else}}
-    use aptos_framework::fungible_asset_u256::FungibleAsset;
-    {{/if}}
+    use aptos_framework::fungible_asset_u256::
+    {{~#if has_eth_token_metadata}}{FungibleAsset, zero}{{else}}FungibleAsset{{/if}};
     {{else if has_eth_token_metadata}}
     use aptos_framework::fungible_asset_u256::zero;
     {{/if}}


genesis-builder/src/generate.rs
```diff
@@ -18,6 +18,9 @@ struct L2Module {
     structs: BTreeMap<String, Vec<L2Input>>,
     has_fungible_asset: bool,
     has_non_empty_args: bool,
+    has_evm_call: bool,
+    has_evm_view: bool,
```
bulenttastan, line 22: We probably don't need both `has_call` and `has_view`. Generally this is not true, but for L2 contracts one implies the other is missing.

genesis-builder/l2_move_template.hbs
```diff
@@ -1,9 +1,16 @@
 module {{name}}::{{snake name}} {
-    use aptos_framework::fungible_asset_u256::
-    {{~#if has_fungible_asset}}{FungibleAsset, zero}{{else}}zero{{/if}};
-    use EthToken::eth_token::get_metadata;
+    {{#if has_fungible_asset}}
```
bulenttastan, line 2: Can we use `{{~if}}` here with only `FungibleAsset` and `zero`? This will append it to the previous line.

bulenttastan (approved):
Left some comments on the handlebar code. The rest looks great.
Approving as these are small changes, merging as is also fine.

genesis-builder/src/generate.rs
birchmd: You're right, all the L2 contracts have view methods so `has_evm_view` is not needed. `has_evm_call` is needed because `L1BlockNumber.move` only contains view functions.

Fixed in https://github.com/UmiNetwork/op-move/pull/358/commits/694d9e51463aec8a8cccd577141f46c3c2048918

genesis-builder/l2_move_template.hbs
birchmd: Done in https://github.com/UmiNetwork/op-move/pull/358/commits/157385c50eb8dc51c7e2486648193597b9f01352

bulenttastan (approved):

genesis-builder/src/generate.rs
```diff
@@ -84,9 +93,21 @@ pub fn l2_abi_to_move() -> anyhow::Result<()> {
                     ..Default::default()
                 };
 
-                if fun.state_mutability == StateMutability::Payable {
-                    function.has_value = true;
-                    has_fungible_asset = true;
+                match &fun.state_mutability {
+                    StateMutability::Payable => {
+                        has_evm_call = true;
+                        function.is_view = false;
```
ilchu, line 99: I believe if it's instantiated with `Default::default()`, only flipping to `true` is needed to be explicit. But it does aid readability, so mostly a nit.

ilchu (approved):
Really like seeing the native impls seeing functionality reuse!

genesis-builder/src/generate.rs
birchmd: Yes, I like the explicit setting to `false` even though it's not strictly necessary. It may avoid problems in the future if we somehow accidentally change the behaviour of Default; it's quite unlikely of course, but the main point is that unchecked assumptions (such as assuming a `function` is marked as "not view" by default) can lead to bugs in future code changes, so it's safer to minimize the number of assumptions.

Issue #247, opened 2025-01-30
State storage failures should surface up


bulenttastan:
In-memory storage failures is not as expected, so we forced state tree access with `.expect()`, however this will become an issue when the tree is accessed from RocksDB. Persistent DB failures are more expected and should be handle properly.

https://github.com/MovedNetwork/op-move/blob/5bff3017103056dbc9c1b2d4a300fc22089fd81e/state/src/lib.rs#L20

birchmd:
Done in #387

PR #359, opened 2025-05-15
perf: Create separate owned `ApplicationReader` to allow for lock-free command and query processing

code-dogg:
### Description
Removes the lock wrapping `Application`, allowing for completely lock-free independent reads and writes. This increases throughput while introducing eventual consistency.

### Changes
* Remove `Arc<RwLock<...>>` wrapper surrounding `Application` and replace it with a single mutable reference.
* Use scoped futures instead of static futures for running `CommandActor` queue `work` processing.
* Introduce `ApplicationReader` that implements all queries and is independent of `Application`.
* Use `StateQueries` instead of `State` for query processing
* Implement `StateQueries::resolve_at` for getting resolver at current or any historical block height

### Notes
Currently, the background worker for command queue runs in a scoped thread to make it possible to run threads without `'static` lifetime. The reason is that `CommandActor` holds a mutable reference to `Application`. The scoping mechanism is analogous to `std` scoped threads and uses `tokio_scoped` crate to achieve the same thing for `async` futures.

#### Todo
* ~~Use references to `Application` and `ApplicationReader` instead of owned values~~
* ~~Fix benchmark code that now cannot use `Arc` owned value, might be fixed by the above~~
* ~~Investigate the usage of `tempfile` for lmdb storage in tests, the directory will not be cleaned on shutdown~~
* ~~The op stack calls `engine_getPayload` before the payload is ready, this is not a problem necessarily, but it makes it repeat the forkchoice calls which may slow down block production. Possible fix would be to call `queue.wait_for_pending_commands` inside `getPayload` RPC handler.~~

### Benchmark
violin

From the benchmark above, we can observe that the **throughput** for blocks with a single Optimism system transaction at least **tripled**.


Cargo.toml
```diff
@@ -96,8 +96,10 @@ sui-types = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.28.3"
 test-case = "3"
 thiserror = "1"
 tokio = { version = "1", features = ["full"] }
+tokio-scoped = "0.2.0"
```
birchmd, line 99: I am hesitant to use this crate because it is archived on GitHub, so it must not be maintained anymore. Its README suggests `async_scoped` instead, although that crate is [not recommended for use in production code](https://github.com/rmanoka/async-scoped/issues/33). That issue suggests to use `Arc` instead of scoping. So I wonder if we should only remove the `RwLock` and leave `Arc`.

birchmd (commented):
I haven't done anywhere near a full review yet, just wanted to start with researching the newly added dependency. I'll do a thorough review of the rest of the changes, but I want to start the discussion of `tokio-scoped` right away since it could lead to changing this PR.

Cargo.toml
code-dogg: Thanks for pointing that out. I realized that since the code is structured so that it always goes through `moved_app::run` with two futures to run concurrently, there is no need to call any `spawn`, but a simple `tokio::join` would do. It improved the performance, simplified the code and got rid of the dependency.

birchmd: Awesome! Good catch!

app/src/command.rs
```diff
@@ -32,6 +32,15 @@ use {
 
 impl<D: Dependencies> Application<D> {
     pub fn start_block_build(&mut self, attributes: Payload, id: PayloadId) {
+        if self
+            .payload_queries
+            .by_id(&self.storage_reader, id)
+            .unwrap()
```
birchmd, line 38: Since this is just an early exit, we could avoid the `unwrap` here with `.ok().flatten()`. Although there are other unwraps in the function from storage access, so probably this function should just be refactored to return a `Result`. It can be done in another PR, probably as part of the work for #247

storage/heed/src/state.rs
```diff
@@ -138,10 +142,14 @@ impl<'db> HeedStateQueries<'db> {
 
         transaction.commit()?;
 
-        Ok(height?.unwrap_or(0))
+        Ok(height?.unwrap_or(0).max(1))
```
birchmd, line 145: Why is 1 the minimum allowed height to be returned?

storage/rocksdb/src/state.rs
```diff
@@ -122,29 +126,35 @@ impl<'db> RocksDbStateQueries<'db> {
             .db
             .get_pinned_cf(&self.height_cf(), HEIGHT_KEY)?
             .map(|v| u64::from_key(v.as_ref()))
-            .unwrap_or(0))
+            .unwrap_or(0)
+            .max(1))
```
birchmd, line 130: Why is 1 the minimum allowed height to be returned?

server/src/dependency/in_memory.rs
```diff
@@ -1,94 +1,157 @@
...
+pub struct InMemoryDependencies {
+    memory_reader: moved_blockchain::in_memory::SharedMemoryReader,
+    memory: Option<moved_blockchain::in_memory::SharedMemory>,
+    receipt_memory_reader: moved_blockchain::receipt::ReceiptMemoryReader,
+    receipt_memory: Option<moved_blockchain::receipt::ReceiptMemory>,
+    trie_db: Arc<moved_state::InMemoryTrieDb>,
+}
+
+impl InMemoryDependencies {
+    pub fn new() -> Self {
```
birchmd, line 34: Now that `InMemoryDependencies` has a `new()` method it should also have

```rust
impl Default for InMemoryDependencies {
    fn default() -> Self {
        Self::new()
    }
}
```

I'm surprised clippy didn't complain about it.

app/src/dependency.rs
```diff
@@ -1,13 +1,61 @@
...
+    pub payload_queries: D::PayloadQueries,
+    pub receipt_queries: D::ReceiptQueries,
+    pub receipt_memory: D::ReceiptStorageReader,
+    pub storage: D::SharedStorageReader,
+    pub state_queries: D::StateQueries,
+    pub evm_storage: D::StorageTrieRepository,
+    pub transaction_queries: D::TransactionQueries,
+}
+
+unsafe impl<D: Dependencies> Sync for ApplicationReader<D> {}
```
birchmd, line 23: I suggest adding the annotation

```rust
#[cfg(any(feature = "test-doubles", test))]
```

We do not want to accidentally make the production code unsound through unsafe code. Even requiring this for the tests to compile is a bit of a red flag which probably deserves some investigation.

birchmd (commented):
Nice work on this massive PR. It's great to have these performance improvements! There are a few things I think need to be addressed before we merge this (spawning the web server task and annotating the unsafe code as test-only) as well as a few nits, but otherwise it looks good!

server/src/tests/payload.rs
```diff
@@ -0,0 +1,111 @@
...
+    moved_api::schema::{
+        BlobsBundleV1, ExecutionPayloadV3, ForkchoiceUpdatedResponseV1, GetPayloadResponseV3,
+        PayloadId, PayloadStatusV1, Status,
+    },
+    moved_execution::U256,
+    moved_shared::primitives::{Address, Bytes, B2048, B256, U64},
+};
+
+#[tokio::test]
+async fn test_sending_the_same_payload_twice_produces_one_block() -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
code-dogg: I think this behavior should be documented explicitly in the form of a test

storage/heed/src/state.rs
code-dogg: Because the `height` at zero is always filled in through the constructor.

On second thought, this is not really expected behavior. It should return current height.

storage/rocksdb/src/state.rs
code-dogg: :arrow_up:

server/src/tests/test_context.rs
```diff
@@ -220,3 +181,36 @@ pub async fn handle_request<T: DeserializeOwned>(
...
+                continue;
+            }
+        }
+
+        let result: T =
+            serde_json::from_value(response.result.expect("If not error then has result"))?;
+        return Ok(result);
+    }
+
+    unreachable!()
```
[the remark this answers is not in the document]
code-dogg: The `error` is not in scope at that spot, but we can use `loop` like a do-while.

server/src/lib.rs
```diff
@@ -77,68 +79,86 @@ pub async fn run(max_buffered_commands: u32, max_concurrent_queries: u32) {
...
-        .with(warp::cors().allow_any_origin());
+    run_with_app(queue, state, app_reader).await;
+}
+
+async fn run_with_app(
+    queue: CommandQueue,
+    state: CommandActor<'_, dependency::Dependency>,
+    app_reader: ApplicationReader<dependency::Dependency>,
+) {
+    moved_app::run(state, async move {
```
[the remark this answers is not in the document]
code-dogg: Great suggestion

app/src/dependency.rs
code-dogg: I don't like this either, it should be possible to get rid of this. It's because the `ReadHandle` of `evmap` requires that every thread has it's own copy, but currently the copy is created from a reference, that is being sent between threads. Sending the reference over thread boundaries is not a problem as long as it is only used for cloning. But due to this property it manually does not implement `Sync`. The way to get rid of this unsafe impl would be to have the `ApplicationReader` owned handle being created from something else that is `Sync`. In `evmap` the `ReadHandleFactory` exists for this purpose, but I did not use it in the implementation that would likely require further restructuring.

code-dogg: Running the binary with in-memory backend is still useful for debugging though so I would rather not add this `cfg` marker.

Cargo.toml
```diff
@@ -98,6 +98,7 @@ thiserror = "1"
 tokio = { version = "1", features = ["full"] }
 warp = "0.3"
 warp-reverse-proxy = "1"
+evmap = "10"
```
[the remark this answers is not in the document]
code-dogg: Since this does not affect production code and is a problem you would rarely encounter, I suggest we address this separately.

birchmd (approved):

app/src/dependency.rs
birchmd: Fair enough. I think it should be relatively high priority to replace `evmap` and remove this unsafe impl in a follow up PR.

PR #361, opened 2025-05-15
Feat(tx): Allow transactions to execute EVM contracts directly

birchmd:
### Description
Allows users to construct a transaction which calls the EVM native directly. This simplifies the flow for calling an EVM contract because users do not need to worry about constructing the entry function object that was needed previously (see for example the simplification of the `call_contract` function in `server/src/tests/evm_contracts.rs`).

### Changes
- Change the byhaviour of `system_evm_call` to take payment for the `value` from the user, eliminating the need for a separate `burn_eth` operation after execution an EVM transaction. It also automatically publishes logs from the EVM and aborts the transaction if the EVM fails (so that paying the value is atomic with the success of the EVM).
- New `system_deposit_evm_call` which is used only in the deposited transaction flow. It has the old behaviour of `syestem_evm_call` where it just calls the EVM native directly.
- New `TransactionData` variant to allow calling EVM contracts directly instead of needed to construct the entry function for the EVM native.

### Testing
- Existing test (one updated to use the new EVM contract transaction type)
- Integration test


bulenttastan (approved):
Good work. It's nice to remove the potentially insecure `burn` method.
@ilchu please take a look too for any suggestions.

execution/src/transaction.rs
```diff
@@ -368,6 +384,41 @@ impl From<TransactionRequest> for NormalizedEthTransaction {
...
+// Intentionally left private to hide the serialization details
+// from users of `TransactionData`. This allows making changes to
+// `TransactionData` itself while remaining backwards compatible with
+// the serialization format.
+// The purpose of `Cow` wrapping all the data is to allow serializing
+// from a reference to `TransactionData` without cloning while also allowing
+// deserializing to `SerializableTransactionData` with owned data.
+// Data type which are `Copy` are left without `Cow` references.
+#[derive(Deserialize, Serialize)]
+enum SerializableTransactionData<'a> {
```
ilchu, line 396: For versioning, we could also keep separate private submodules in the future. And it would also allow us flexibility in ser/de implementations if we'd need them custom. I was trying to think of a leaner separation that we could have, but I guess most zero-copy approaches or even a simple `#[serde(borrow)]` would only take care of the deserialization part, without clone on write. And it's much more pleasant when the main public type doesn't have the lifetime generics.

ilchu (approved):
This was a much lighter change than I expected. Great stuff!

execution/src/simulate.rs
```diff
@@ -144,6 +146,32 @@ pub fn call_transaction(
...
+                tx.value,
+                tx.data.to_vec(),
+                verify_input.session,
+                verify_input.traversal_context,
+                verify_input.gas_meter,
+                &code_storage,
+            )?;
+            Ok(outcome.output)
+        }
+        TransactionData::EvmContract { address, data } => {
```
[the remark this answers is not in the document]
bulenttastan: There might be other reasons, but one reason is that `op-node` will not prepend the bytecode with the enum selection, so we have to accept it as pure EVM call.

execution/src/transaction.rs
birchmd: Yes it might make sense to factor it out into a separate module in the future. Probably next time we make a change to it we should do that.

PR #362, opened 2025-05-21
fix(app): ensure per account nonce ordering in the mempool

ilchu:
### Description

While trying to generate a block with many transactions at once for a test in the fee history PR, I realized that multiple transactions were not executed successfully, and it was relatively random. Debugging showed it was due to draining our mempool which had an underlying Hashmap implementation, and it didn't preserve per account nonce ordering even within a single block, so this PR fixes that. 

### Changes

- A slightly more involved custom type for the mempool, mostly inspired by Aptos
- Slight rework of the transactions passed around in the execution loop

### Testing

- [x] New test for this case in the app crate

### Notes

This would be a blocker for finishing the tests in the fee history PR, so this needs to be merged first. Also there are a few additional features that we could have in relation to the mempool, such as one more map for faster nonce lookups or tx invalidation, but I tried to keep it minimal for now.


birchmd (approved):
Good find! And very nice to see the mempool logic being nicely encapsulated now.

In the future we may think about adding some crash resilience to the mempool as well. Although it might not be needed if we simply replay the `eth_sendRawTransaction` requests from the server in the case of a crash.

app/src/mempool.rs
```diff
@@ -0,0 +1,92 @@
+use std::collections::{BTreeMap, HashMap};
```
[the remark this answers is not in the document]
ilchu: Didn't notice it as CI was happy. I lurked around formatting options and it seems to be governed by [this](https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=#imports_granularity). Do we want to have a `rustfmt.toml` and enforce it automatically?

app/src/mempool.rs
```diff
@@ -0,0 +1,92 @@
...
+
+        account_txs.insert(value.tx_nonce, value)
+    }
+
+    /// Drains all transactions from the [`Mempool`], returning them in a sensible order
+    /// for block inclusion (ordered by account, then by nonce).
+    pub fn drain(&mut self) -> impl Iterator<Item = MempoolTransaction> {
+        let txs = std::mem::take(&mut self.txs);
+
+        // TODO: any possible dependencies between accounts to be taken care of?
```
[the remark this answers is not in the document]
ilchu: I'll have it removed then.

bulenttastan (approved):
Very nice 🙌

app/src/mempool.rs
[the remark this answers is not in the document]
birchmd: We do have it enforced in [rustfmt.toml](https://github.com/UmiNetwork/op-move/blob/6069f3d1edb879bbeefd71eaba3c62c0fdab8281/rustfmt.toml#L1), so not sure why CI didn't catch it. I think that formatting option only works on nightly Rust, so maybe CI is only running stable Rust currently?

birchmd (approved):

app/src/mempool.rs
```diff
@@ -0,0 +1,90 @@
...
+                .recover_signer()
+                .map_err(|_| InvalidTransactionCause::InvalidSigner.into()),
+            OpTxEnvelope::Eip2930(signed) => signed
+                .recover_signer()
+                .map_err(|_| InvalidTransactionCause::InvalidSigner.into()),
+            OpTxEnvelope::Eip1559(signed) => signed
+                .recover_signer()
+                .map_err(|_| InvalidTransactionCause::InvalidSigner.into()),
+            OpTxEnvelope::Eip7702(_) | OpTxEnvelope::Deposit(_) => Err(Error::InvariantViolation(
+                moved_shared::error::InvariantViolation::MempoolTransaction,
```
code-dogg, line 50: Is this really an invariant violation? Is it impossible for the user to submit Eip7702 or a Deposit transaction?

app/src/mempool.rs
```diff
@@ -0,0 +1,90 @@
...
+    op_alloy::consensus::OpTxEnvelope,
+    std::collections::{BTreeMap, HashMap},
+};
+
+type Nonce = u64;
+
+#[derive(Debug, Clone)]
+pub struct MempoolTransaction {
+    pub inner: OpTxEnvelope,
+    // Nonce taken out of the envelope for easier ordering
```
code-dogg, line 18: Maybe convert this into a doc comment `///`?

app/src/command.rs
```diff
@@ -175,13 +174,14 @@ impl<D: Dependencies> Application<D> {
         // https://github.com/ethereum-optimism/specs/blob/9dbc6b0/specs/protocol/deposits.md#kinds-of-deposited-transactions
         let l1_fee = transactions
             .peek()
-            .and_then(|(_, v, _)| v.as_deposit())
+            .and_then(|tx| tx.inner.as_deposit())
```
code-dogg, line 177: `MempoolTransaction` could have this method, so that you don't need to expose it's insides and rather it can have an API.

app/src/command.rs
```diff
@@ -144,14 +146,11 @@ impl<D: Dependencies> Application<D> {
...
-            tx_hash,
-            (
-                OpTxEnvelope::try_from_eth_envelope(tx)
-                    .unwrap_or_else(|_| unreachable!("EIP-4844 not supported")),
-                encoded,
-            ),
-        );
+        let nonce = tx.nonce();
+        let inner = OpTxEnvelope::try_from_eth_envelope(tx)
+            .unwrap_or_else(|_| unreachable!("EIP-4844 not supported"));
```
code-dogg, line 151: I know this was here before, but is it really unreachable?

app/src/mempool.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    /// Drains all transactions from the [`Mempool`], returning them in a sensible order
+    /// for block inclusion (ordered by account, then by nonce).
+    pub fn drain(&mut self) -> impl Iterator<Item = MempoolTransaction> {
+        let txs = std::mem::take(&mut self.txs);
+
+        txs.into_iter()
+            .flat_map(|(_, account_txs)| account_txs.into_values())
+    }
+}
```
code-dogg, line 90: Seems like a nice place for some tests.

code-dogg (commented):
Great I like how you identified the problem and made an elegant solution while using encapsulating it in the `Mempool` service.

All my comments are nits expect for the invariant violation, I would like that reviewed by you to maintain the error meanings.

app/src/mempool.rs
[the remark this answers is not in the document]
bulenttastan: Maybe the github action is not working as intended, cause otherwise it should be picking up the nightly toolchain:
[.github/workflows/format_check.yaml](https://github.com/UmiNetwork/op-move/blob/main/.github/workflows/format_check.yaml)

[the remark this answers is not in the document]
ilchu: Indeed, that's kind of weird 🤔

app/src/mempool.rs
ilchu: It's a bit convoluted, but the only entry point into the mempool is through `add_transaction()` call, which takes a `TxEnvelope`, so that excludes deposited txs. And even if a EIP-7702 was submitted, it would fail at normalization stage, though truth be told it fails with an `InvalidTransactionCause`. So I guess I'll make the `signer()` function private and change to invalid transaction error instead. Thanks for pointing that out! Though there still remains a bit of discrepancy where we don't allow deposit txs *into* the mempool, but we do cast all the payload-submitted txs into `MempoolTransaction` for uniformity, but it never gets a signer method called.

birchmd: It is a security invariant that users cannot submit `Deposit` transactions. Otherwise users could mint tokens for free!

EIP-7702 is just not something we currently support and probably we will not ever support it. It enables EVM-based account abstraction, but since we expect all our EOAs (externally owned addresses) to be Move accounts, account abstraction will follow different patterns on our system.

app/src/command.rs
birchmd: Along similar lines of EIP-7702, we don't support EIP-4844 because it is a specific transaction kind for submitting "blobs" (most often rollup data) to EVM. Supporting this standard would allow someone to build an "L3" rollup on top of our L2 rollup. But I don't see any use case for that in the short to medium term and so I don't expect us to support it for a long time, if ever.

It is a good point though that having `unreachable`s is a bit of a code smell. The best way to avoid this would be to have our own types which only has the variants we use. But using the alloy types is beneficial because we have high confidence they are correct and it reduces the amount of code we have to maintain ourselves. Maybe there is a way to get the best of both worlds with some kind of new type wrapper.

An alternative would be to use something like [verus](https://github.com/verus-lang/verus) to have proofs that all the `unreachable`s truly are unreachable in all code paths.

ilchu: It would fail at normalization stage, just like EIP-7702. And we don't support blobs in many other places, such as fee history or gas calculations. But you're right that it can be triggered indeed, as it's an externally facing API. I do feel it's better to have a separate issue for this, though. One idea I had is that maybe we could force the tx normalization to be happening earlier, right when a tx is added to mempool or parsed from payload. That would essentially make it a pending / commited tx flow.

birchmd: Yes, I agree handling this should be a separate issue. We should reject unsupported transaction types by returning an error in the `eth_submitRawTransaction` endpoint (this is the only way users can add transactions to the system).

app/src/command.rs
ilchu: Considering that there's some juggling of the inner `OpTxEnvelope` down the line and that we're not trying to enforce any invariants here or futureproof things, I believe it's fine to just call the inner methods directly. Of course we can change it when the pending tx functionality expands.

app/src/command.rs
ilchu: Created #365.

PR #363, opened 2025-05-21
Feat(tx): charge gas for Move module deployment

### LINKED ISSUE #346: Charge gas for module deployment

birchmd:
### Description
Fixes #346 

### Changes
- Charge gas for module deployment

### Testing
- Test for `eth_estimateGas` shows increased gas cost.
- I tried writing a dedicated test, but isolating the deploy gas cost from the intrinsic transaction gas cost ended up being too much of a pain.

### Notes
This PR assumes the gas cost used in the Aptos native for publishing code is sufficient for our purposes too. I think it should be, though I am not fully certain.


api/src/methods/estimate_gas.rs
```diff
@@ -140,7 +140,7 @@ mod tests {
         drop(state_channel);
         state_handle.await.unwrap();
 
-        let expected_response: serde_json::Value = serde_json::from_str(r#""0x6326""#).unwrap();
+        let expected_response: serde_json::Value = serde_json::from_str(r#""0x6330""#).unwrap();
```
bulenttastan, line 143: nit: it goes up by 4 for the deployment, just seems suspiciously low.

bulenttastan (approved):

api/src/methods/estimate_gas.rs
birchmd: Yes, good call. I dug into it more and found that we are missing charging on a whole type of gas (storage costs). I created a new issue for it: #364 . I think we should merge this PR because it still does add a necessary cost. The storage costs will be handled in another PR.

code-dogg (approved):
Great catch on the storage charge

PR #375, opened 2025-05-30
feat(app): ring buffer for the `BLOCKHASH` opcode

### LINKED ISSUE #352: Efficient implementation of EVM `BLOCKHASH` opcode
The EVM has an opcode which an retrieve the hash of a block given a height, but requires that height to be within the 256 most recent blocks. Currently there is a "dumb" implementation that queries the database for an entire block only to return its hash. This is certainly too slow for the amount of gas the EVM changes for this opcode (it is less than the cost of a keccak256 hash for example). To have the gas cost align with the real resource usage we essentially need to reduce this to an in-memory lookup. The purpose of this issue is to build the implementation that maintains the most recent 256 block hashes in memory (using a ring buffer for example, `VecDeque` from the standard library should work fine) and uses this structure to serve the EVM opcode.

In the context of #268 this in-memory structure will need to be constructed from the persistent storage at startup.

Code reference: https://github.com/MovedNetwork/op-move/blob/408859774ec1ae534c7db80dbe0ebeb635b72214/app/src/block_hash.rs#L23

ilchu:
### Description

Replaces a db lookup for the BLOCKHASH opcode with a ring buffer implementation.

Fixes #352.

### Changes

- Added the buffer into dependencies
- VecDeque-based implementation with a thread safe wrapper

### Testing

Some new trivial tests, all old ones (with `evm_contracts.rs` being of special interest) pass too.

### DIFF
--- a/app/src/block_hash.rs
+++ b/app/src/block_hash.rs
@@ -1,37 +1,226 @@
 use {
-    alloy::primitives::B256, umi_blockchain::block::BlockQueries,
-    umi_evm_ext::state::BlockHashLookup,
+    alloy::primitives::B256,
+    std::{
+        collections::VecDeque,
+        sync::{Arc, RwLock},
+    },
+    umi_blockchain::block::BlockQueries,
+    umi_evm_ext::state::{BlockHashLookup, BlockHashWriter},
 };
 
-pub struct StorageBasedProvider<'a, S, B> {
-    storage: &'a S,
-    block_query: &'a B,
+const BLOCKHASH_HISTORY_SIZE: usize = 256;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct BlockHashEntry {
+    number: u64,
+    hash: B256,
 }
 
-impl<'a, S, B> StorageBasedProvider<'a, S, B>
-where
-    B: BlockQueries<Storage = S>,
-{
-    pub fn new(storage: &'a S, block_query: &'a B) -> Self {
+#[derive(Debug, Clone)]
+pub struct BlockHashRingBuffer {
+    entries: VecDeque<BlockHashEntry>,
+    /// Block number of the latest block stored for validation purposes
+    latest_block: u64,
+}
+
+impl Default for BlockHashRingBuffer {
+    fn default() -> Self {
         Self {
-            storage,
-            block_query,
+            entries: VecDeque::with_capacity(BLOCKHASH_HISTORY_SIZE),
+            latest_block: 0,
         }
     }
 }
 
-// TODO: looking up the entire block just to get its hash is inefficient;
-// especially since we do not need the whole history, only the most recent 256
-// block hashes. This implementation should be replaced with something better.
-impl<S, B> BlockHashLookup for StorageBasedProvider<'_, S, B>
-where
-    B: BlockQueries<Storage = S>,
-{
+impl BlockHashRingBuffer {
+    pub fn push(&mut self, block_number: u64, block_hash: B256) {
+        // Ensure blocks are added in sequence (detect potential reorgs)
+        if block_number > 0 && self.latest_block > 0 && block_number != self.latest_block + 1 {
+            eprintln!(
+                "WARN: Block hash buffer - expected block {}, got {}",
+                self.latest_block + 1,
+                block_number
+            );
+        }
+
+        self.entries.push_back(BlockHashEntry {
+            number: block_number,
+            hash: block_hash,
+        });
+
+        while self.entries.len() > BLOCKHASH_HISTORY_SIZE {
+            self.entries.pop_front();
+        }
+
+        self.latest_block = block_number;
+    }
+
+    // TODO: make this a builder method?
+    pub fn initialize_from_storage<S, B>(&mut self, storage: &S, block_query: &B, latest_block: u64)
+    where
+        B: BlockQueries<Storage = S>,
+    {
+        self.entries.clear();
+        self.latest_block = 0;
+
+        let start_block = if latest_block >= BLOCKHASH_HISTORY_SIZE as u64 {
+            latest_block - BLOCKHASH_HISTORY_SIZE as u64 + 1
+        } else {
+            0
+        };
+
+        for block_num in start_block..=latest_block {
+            if let Ok(Some(block)) = block_query.by_height(storage, block_num, false) {
+                self.push(block_num, block.0.header.hash);
+            }
+        }
+    }
+}
+
+impl BlockHashLookup for BlockHashRingBuffer {
+    fn hash_by_number(&self, block_number: u64) -> Option<B256> {
+        if block_number > self.latest_block {
+            // Future block
+            return None;
+        }
+
+        let blocks_ago = self.latest_block - block_number;
+        if blocks_ago > BLOCKHASH_HISTORY_SIZE as u64 {
+            // Too old
+            return None;
+        }
+
+        for entry in self.entries.iter().rev() {
+            if entry.number == block_number {
+                return Some(entry.hash);
+            }
+            // Since we're going backwards chronologically, if we've passed the target block, it's not there
+            if entry.number < block_number {
+                break;
+            }
+        }
+
+        None
+    }
+}
+
+impl BlockHashWriter for BlockHashRingBuffer {
+    fn push(&mut self, height: u64, hash: B256) {
+        self.push(height, hash);
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct SharedBlockHashCache {
+    inner: Arc<RwLock<BlockHashRingBuffer>>,
+}
+
+impl SharedBlockHashCache {
+    pub fn initialize_from_storage<S, B>(&mut self, storage: &S, block_query: &B, latest_block: u64)
+    where
+        B: BlockQueries<Storage = S>,
+    {
+        if let Ok(mut cache) = self.inner.write() {
+            cache.initialize_from_storage(storage, block_query, latest_block);
+        }
+    }
+}
+
+impl Default for SharedBlockHashCache {
+    fn default() -> Self {
+        Self {
+            inner: Arc::new(RwLock::new(BlockHashRingBuffer::default())),
+        }
+    }
+}
+
+impl BlockHashLookup for SharedBlockHashCache {
     fn hash_by_number(&self, number: u64) -> Option<B256> {
-        let block = self
-            .block_query
-            .by_height(self.storage, number, false)
-            .ok()??;
-        Some(block.0.header.hash)
+        if let Ok(cache) = self.inner.read() {
+            cache.hash_by_number(number)
+        } else {
+            // TODO: deal with poisoning?
+            None
+        }
+    }
+}
+
+impl BlockHashWriter for SharedBlockHashCache {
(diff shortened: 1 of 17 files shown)


app/src/block_hash.rs
```diff
@@ -1,37 +1,226 @@
...
-// block hashes. This implementation should be replaced with something better.
-impl<S, B> BlockHashLookup for StorageBasedProvider<'_, S, B>
-where
-    B: BlockQueries<Storage = S>,
-{
+impl BlockHashRingBuffer {
+    pub fn push(&mut self, block_number: u64, block_hash: B256) {
+        // Ensure blocks are added in sequence (detect potential reorgs)
+        if block_number > 0 && self.latest_block > 0 && block_number != self.latest_block + 1 {
+            eprintln!(
```
bulenttastan, line 39: Should this stop the server?

app/src/block_hash.rs
```diff
@@ -1,37 +1,226 @@
...
+        if blocks_ago > BLOCKHASH_HISTORY_SIZE as u64 {
+            // Too old
+            return None;
+        }
+
+        for entry in self.entries.iter().rev() {
+            if entry.number == block_number {
+                return Some(entry.hash);
+            }
+            // Since we're going backwards chronologically, if we've passed the target block, it's not there
```
bulenttastan, line 97: Can this happen if blocks are always in sequence?

app/src/block_hash.rs
```diff
@@ -1,37 +1,226 @@
...
+        // But let's test with a more realistic scenario
+        for i in 102..400u64 {
+            buffer.push(i, B256::from([(i % 256) as u8; 32]));
+        }
+
+        // Now current block is 399
+        // Block 100 should be too old (399 - 100 = 299 > 256)
+        assert_eq!(buffer.hash_by_number(100), None);
+
+        // Block 143 should still be accessible (399 - 143 = 256, which is the limit)
```
bulenttastan, line 220: Let's make a single comment saying it's not accessible because the limit should be less than 256. It's a strange comment saying it's accessible and then in the next line we say, oops just kidding.

evm-ext/src/state/block_hash.rs
```diff
@@ -6,6 +6,11 @@ pub trait BlockHashLookup {
     fn hash_by_number(&self, number: u64) -> Option<B256>;
 }
 
+/// A trait for keeping block hashes for 0x40 retrieval
```
bulenttastan, line 9: We should make what 0x40 means clearer `0x40 opcode blockhash retrieval`

app/src/dependency.rs
```diff
@@ -122,6 +127,8 @@ pub trait DependenciesThreadSafe:
     Dependencies<
         BaseTokenAccounts: Send + 'static,
         BlockHash: Send + 'static,
+        BlockHashLookup: Send + 'static,
+        BlockHashWriter: Send + 'static,
```
bulenttastan, line 131: Should we separate lookup and writer even though they refer to the same cache?

app/src/block_hash.rs
ilchu: I feel like no, because the block hash functionality is just there for the speedup and nothing critical happens if the order is broken, though that's definitely an abnormality. I admit it would make sense to e.g. have some fallback to block queries as we were doing before.

app/src/block_hash.rs
```diff
@@ -1,37 +1,226 @@
...
+        });
+
+        while self.entries.len() > BLOCKHASH_HISTORY_SIZE {
+            self.entries.pop_front();
+        }
+
+        self.latest_block = block_number;
+    }
+
+    // TODO: make this a builder method?
```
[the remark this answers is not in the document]
ilchu: My concern here was that the mutation of an existing instance with full wipe of internal structures could be confusing. If it's meant for initialization, then it should be just that, an alternative constructor not taking the `self` argument.

app/src/block_hash.rs
ilchu: This loops back to that `eprintln!` message [above](https://github.com/UmiNetwork/op-move/pull/375#discussion_r2117159789), as that's the only case where it could happen in a running application. Though considering that it's meant to be initialized from storage at startup, I can also see a scenario where there's some uncaught data corruption happening that jumbled up this order.

server/src/tests/evm_contracts.rs
```diff
@@ -20,6 +20,7 @@ mod evm_contract {
         // This contract has one function which uses the `BLOCKHASH` EVM opcode
         // to try to get the hash for block number 3. It emits the response
         // as an event.
+    #[derive(Debug)]
```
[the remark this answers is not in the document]
bulenttastan: Interesting and actually it's pretty cool. We should still indent 1 tab to right.

code-dogg:
Cache is a layer between storage and execution that offers faster lookup of some data and can hit or miss. On cache hit, the call early returns data from the cache, on cache miss the call is propagated further to the storage layer.

I think here the expectation is that the "cache" will be always consistent and is the primary source of the data. So it's more like a separately built index.

birchmd (commented):
Good work on this. I think there is a little more to do before this can be merged.

app/src/block_hash.rs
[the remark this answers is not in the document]
birchmd: I think it makes sense to be a constructor instead of something that mutates.

evm-ext/src/state/block_hash.rs
```diff
@@ -6,6 +6,11 @@ pub trait BlockHashLookup {
     fn hash_by_number(&self, number: u64) -> Option<B256>;
 }
 
+/// A trait for 0x40 BLOCKHASH opcode retrieval
+pub trait BlockHashWriter: BlockHashLookup {
```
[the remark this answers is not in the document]
ilchu: It wasn't, but decoupling it made for quite a bit more of the boilerplate, as the block lookup struct had to be duplicated in the app struct itself. But I guess that's not too bad if on the flip side concerns are separated more clearly.

app/src/block_hash.rs
```diff
@@ -1,37 +1,224 @@
...
+            entries: VecDeque::with_capacity(BLOCKHASH_HISTORY_SIZE),
+            latest_block: 0,
         }
     }
 }
 
-// TODO: looking up the entire block just to get its hash is inefficient;
-// especially since we do not need the whole history, only the most recent 256
-// block hashes. This implementation should be replaced with something better.
-impl<S, B> BlockHashLookup for StorageBasedProvider<'_, S, B>
```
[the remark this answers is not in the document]
ilchu: This got me thinking too, and seems that's not the case, though I'm still not sure even after looking at REVM code.

We explicitly use the `block_hash_lookup` exclusively for the `block_hash_ref()` method from REVM, which is only triggered when an actual BLOCKHASH opcode is encountered inside the native context. The actual state resolution is done by us before entering those calls, and that's what's in charge of what the end user sees AFAICT. So the only use case for this cache is when the opcode is triggered directly by some contract, like we do in `test_evm_contracts`, and that's mostly it?

I also tried modifying the `estimateGas`/`call` tests to produce a lot of blocks and try to execute requests at really old heights, and it all seems to pass no problem, with no execution path touching the BLOCKHASH itself.

I still have doubts because the actual contract that we're checking there is Move-based, so that could be the issue? Of course, I can implement some hybrid solution, but it seems that older block lookup would be a no-op still due to our `block_hash_ref()` implementation.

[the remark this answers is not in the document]
birchmd: Yes, that is correct. This is only used for the `BLOCKHASH` opcode and nothing else. So the way to see the problem I described would be to use `eth_call` to execute an EVM contract using that opcode and specifying an old block in that request.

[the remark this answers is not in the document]
ilchu: I've updated the PR so that the hybrid version is used for the reader, and the non-hybrid for the mutating part. Though the part where I had to duplicate dependencies for the reader part was not pleasant, but I feel that's the cleanest way, even if a bit annoying.

app/src/block_hash.rs
```diff
@@ -1,37 +1,411 @@
...
+
+impl BlockHashLookup for BlockHashRingBuffer {
+    fn hash_by_number(&self, block_number: u64) -> Option<B256> {
+        if block_number > self.latest_block {
+            return None;
+        }
+
+        let blocks_ago = (self.latest_block - block_number) as usize;
+
+        // second condition to guard against cases when the buffer is not full
```
bulenttastan, line 91: nit: capitalize the first word, like `Second`

app/src/block_hash.rs
```diff
@@ -1,37 +1,411 @@
...
+        }
+    }
+}
+
+impl BlockHashLookup for SharedBlockHashCache {
+    fn hash_by_number(&self, number: u64) -> Option<B256> {
+        if let Ok(cache) = self.inner.read() {
+            cache.hash_by_number(number)
+        } else {
+            // TODO: deal with poisoning?
```
bulenttastan, line 140: Is this a real issue? If there is no cache set up yet, then there is no block hash in it, right?

app/src/block_hash.rs
```diff
@@ -1,37 +1,411 @@
...
+impl<S, B> BlockHashLookup for SharedHybridBlockHashCache<'_, S, B>
+where
+    S: Clone,
+    B: Clone + BlockQueries<Storage = S>,
+{
+    fn hash_by_number(&self, number: u64) -> Option<B256> {
+        if let Ok(cache) = self.inner.read() {
+            cache.hash_by_number(number)
+        } else {
+            // TODO: deal with poisoning?
```
bulenttastan, line 233: same here.

evm-ext/src/state/block_hash.rs
```diff
@@ -6,8 +6,17 @@ pub trait BlockHashLookup {
...
 }
 
+/// A trait for 0x40 BLOCKHASH opcode retrieval
+pub trait BlockHashWriter {
+    fn push(&mut self, height: u64, hash: B256);
+}
+
 impl BlockHashLookup for () {
     fn hash_by_number(&self, _number: u64) -> Option<B256> {
         None
```
bulenttastan, line 16: Isn't this supposed to return the block hash?

ilchu: Yes if we're trying to replicate hashmap's entry interface and to have a check if the value was already present in the cache, which sounds like a good idea. But it wraps around to the concern I've outlined [here](https://github.com/UmiNetwork/op-move/pull/375#discussion_r2149718258).

app/src/block_hash.rs
ilchu: Answered above.

app/src/block_hash.rs
```diff
@@ -1,37 +1,411 @@
...
+        }
+
+        self.latest_block = block_number;
+    }
+
+    pub fn initialize_from_storage<S, B>(storage: &S, block_query: &B) -> Self
+    where
+        B: BlockQueries<Storage = S>,
+    {
+        let latest_block = block_query.latest(storage).unwrap().unwrap_or(0);
```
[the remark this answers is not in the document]
ilchu: See [this](https://github.com/UmiNetwork/op-move/pull/375#discussion_r2149718258) comment.

PR #377, opened 2025-06-02
refactor(state): Unify `apply` and `apply_with_tables`

code-dogg:
### Description
Removes `apply_with_tables` and adds table changes to `apply`.

### Changes

- Remove `State::apply_with_tables`
- Add `Changes` for representing a difference between states
- Include `TableChanges` in `Changes`

### Testing
:green_circle: fmt
:green_circle: clippy
:green_circle: test

### DIFF
--- a/state/src/diff.rs
+++ b/state/src/diff.rs
@@ -1,0 +1,126 @@
+use {
+    core::fmt,
+    move_core_types::effects::ChangeSet,
+    move_table_extension::{TableChange, TableChangeSet},
+    std::fmt::{Debug, DebugStruct, Formatter},
+};
+
+/// Represents the difference between two [`State`]s.
+///
+/// The difference can be [applied]. If the difference is between two states `S` and `S'`, then it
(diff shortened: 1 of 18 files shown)


state/src/diff.rs
```diff
@@ -0,0 +1,126 @@
...
+    std::fmt::{Debug, DebugStruct, Formatter},
+};
+
+/// Represents the difference between two [`State`]s.
+///
+/// The difference can be [applied]. If the difference is between two states `S` and `S'`, then it
+/// must be true that `S' := apply(S, diff)`.
+///
+/// `Changes` are usually produced by running a transaction. If `Changes` are produced by running
+/// a transaction on state `S`, then the `S' := apply(S, diff)` represents the state after.
```
ilchu, line 14: It would probably be clearer to unify the name of changes in the explanation. E.g. in the second sentence it is mentioned as `Changes` first, but then as `diff`.

ilchu (approved):

PR #378, opened 2025-06-02
fix(ci): run gas costs test in release mode

ilchu:
### Description

CI seems to be failing more often now on the gas cost test, so I made a separate job to run that test only in release mode. 

### Changes

- new sub job in CI for building/running tests

### DIFF
--- a/.github/workflows/build_and_run_tests.yaml
+++ b/.github/workflows/build_and_run_tests.yaml
@@ -21,5 +21,18 @@ jobs:
           toolchain: stable
       - name: Free up disk space
         run: rm -rf /opt/hostedtoolcache
-      - name: test
+      - name: Test
         run: cargo test -- --skip integration
+
+  flaky:
+    name: flaky
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions-rs/toolchain@v1
+        with:
+          default: true
+          toolchain: stable
+      - name: Run flaky tests
+        run: cargo test -- --ignored
+        continue-on-error: true


.github/workflows/build_and_run_tests.yaml
```diff
@@ -21,5 +21,17 @@ jobs:
           toolchain: stable
       - name: Free up disk space
         run: rm -rf /opt/hostedtoolcache
-      - name: test
-        run: cargo test -- --skip integration
+      - name: Test
+        run: cargo test -- --skip integration --skip solidity_abi::tests::test_gas_costs
+
+  gas_costs_test:
```
birchmd, line 27: As discussed on the call, we won't add this new job yet. Only skip the flaky test for now and I'll work on making it better this week.

birchmd (approved):

bulenttastan:
The unit test time doesn't seem to be increased by much, so it looks like a good push.
You can compare the test time to: https://github.com/UmiNetwork/op-move/actions/runs/15395955818/job/43316669085?pr=380

We can even revert this change: https://github.com/UmiNetwork/op-move/pull/379/files

ilchu:
> The unit test time doesn't seem to be increased by much, so it looks like a good push. You can compare the test time to: [UmiNetwork/op-move/actions/runs/15395955818/job/43316669085?pr=380](https://github.com/UmiNetwork/op-move/actions/runs/15395955818/job/43316669085?pr=380)
> 
> We can even revert this change: [#379 (files)](https://github.com/UmiNetwork/op-move/pull/379/files)

Right, I thought it was 15-ish minutes, but apparently around 25. But I do believe that @code-dogg 's suggestion is reasonable so let's merge it with those changes.

PR #380, opened 2025-06-02
refactor(state): Remove `db` from `State` trait

code-dogg:
### Description
`State` is not responsible for creating its own `eth_trie::DB` since 0130abb3

It also binds `eth_trie` implementation with the trait. The interface should not assume it's implementation.

### Changes
- Remove `State::db`

### Testing
:green_circle: fmt
:green_circle: clippy
:green_circle: test


birchmd (approved):
Good catch!

PR #381, opened 2025-06-03
ci(flaky): Free up disk space before compilation

code-dogg:
### Description
The job currently always fail due to "not enough space on device" error.

Fixes the problem in the same way as for the regular test job.

### Changes
- Add "free up disk space" step to flaky job before compilation

### Testing
:green_circle: Github actions


bulenttastan (approved):
Looks good. We've done this for other github action too.

PR #383, opened 2025-06-04
ci: Merge jobs into one workflow

code-dogg:
### Description
Currently, there are three workflows with one job each. This PR makes it one workflow with three jobs.

While the end result is mostly the same, it makes sense for these jobs to be under one workflow as they are all concerned with one thing which is to validate the source code. Therefore, they should run together.

### Changes
- Move all jobs into one workflow

### Testing
:green_circle: Github actions

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -42,6 +42,7 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --all-features --all-targets
+          name: "Clippy results for rev ${{ github.sha }}"
 
   format:
     name: Format


.github/workflows/build.yaml
```diff
@@ -0,0 +1,57 @@
...
+    permissions: write-all
+    runs-on: ubuntu-latest
+    env:
+      # Make sure CI fails on all warnings, including Clippy lints
+      RUSTFLAGS: "-Dwarnings --cfg tokio_unstable"
+      CARGO_TERM_COLOR: always
+    steps:
+      - uses: actions/checkout@v4
+      - run: rustup component add clippy
+      - uses: actions-rs/clippy-check@v1
```
ilchu, line 41: I believe this causes a second [job](https://github.com/UmiNetwork/op-move/pull/383/checks?check_run_id=43465304323) to appear. Clippy docs [say](https://doc.rust-lang.org/nightly/clippy/continuous_integration/github_actions.html) that no specific action is needed, apparently.

birchmd: Yeah, probably makes sense to refactor this job while we're changing the GH actions.

code-dogg: This is specific to the https://github.com/actions-rs/clippy-check

It creates the second job to post results in a way that it displays summary and annotations.

The name defaults to `clippy` and is overridable, so I named it `Clippy results for rev <git-sha>`. I added the sha because the docs suggest to give it unique name, although not sure if it's necessary for our use-case.

https://github.com/actions-rs/clippy-check?tab=readme-ov-file#inputs

PR #387, opened 2025-06-05
refactor(api): return `Result` instead of `Option` in queries

ilchu:
### Description

Deals away with many unwraps we had in query handling by making the methods return `Result` and treating errors more nicely. This should also presumably fix batcher interaction issues where it was returning an empty JSON response.

### Changes

- Expanded on `state::Error` for new failure modes
- `Option` -> `Result` refactor

### Testing

Old tests still passing, but I plan to write some new tests that specifically trigger failures and undraft this.

### Notes

There are still two points that I don't quite know how to address, both dealing with flattening of errors. One is what occurs at the `state::Error` -> `shared::Error` conversion, as I was conflicted whether db, trie and rlperrors are ok to be conflated together. Another is at the `shared::Error` -> `JsonRpcError` boundary, as it once again mostly flattens the whole thing into block not found code. Maybe the exact codes returned need to be revisited.


api/src/methods/block_number.rs
```diff
@@ -8,7 +11,8 @@ pub async fn execute(
     app: &ApplicationReader<impl Dependencies>,
 ) -> Result<serde_json::Value, JsonRpcError> {
     parse_params_0(request)?;
-    let response = app.block_number();
+    // this is a generic server error code
+    let response = app.block_number().map_err(transaction_error)?;
```
code-dogg, line 15: I would say that this is worse handling of this kind of an error, because in this case the error is invariant violation, not a user error. Invariant violation is an error that should result in panic, because it is for things that should always be true. But with this handling the error bubbles up into a JSON-RPC response, it won't print to stderr (currently we still log all the requests), and keeps the node running. Which would be appropriate if this was a case of, for example, an invalid input.

So I would say if this error should bubble up then it should still result in panic.

```rust
impl From<InvariantViolation> for JsonRpcError {
    fn from(e: InvariantViolation) -> Self {
        panic!("{e}");
    }
}

File: api/src/methods/get_balance.rs [hunk 2]
```diff
@@ -11,7 +11,7 @@ pub async fn execute(
 
     let response = app
         .balance_by_height(address, block_number)
-        .ok_or(JsonRpcError::block_not_found(block_number))?;
+        .map_err(|_| JsonRpcError::block_not_found(block_number))?;
```
code-dogg, line 14: I think our `shared::error::Error` variants should be descriptive enough to allow for `impl From<shared::error::Error> for JsonRpcError`. It would be great to delegate the error conversions there if possible. Then it's unified and in one place.

The `.map_err` is like a manual override of the conversion, useful when the `From` cannot be implemented or it does not fit the use-case. It's frequent usage makes the code kinda spaghettified.

code-dogg (approved):
Great, left some thoughts.

PR #388, opened 2025-06-05
feat(storage-heed): Remove `'static` lifetime from database references

code-dogg:
### Description
The database is implemented as a static lazily initialized variable. This gives the possibility to take `'static` references from it, which is simpler than dealing with explicit sub-static lifetimes. However, it also makes it bound to the process and cannot be reinitialized.

### Changes
- Remove `'static` from `storage-heed`

### Testing
:green_circle: CI


storage/heed/src/evm.rs
```diff
@@ -36,18 +36,18 @@ impl From<heed::Error> for Error {
 
 #[derive(Debug, Clone)]
 pub struct HeedStorageTrieRepository {
-    env: &'static heed::Env,
+    env: Arc<heed::Env>,
```
ilchu, line 39: What's the reasoning behind only storage and eth trie being Arc-wrapped, and all other queries/repos having a `'db` ref? That's the only thing that evades me.

ilchu (approved):

storage/heed/src/evm.rs
code-dogg: https://github.com/UmiNetwork/op-move/pull/388/files#diff-e45ea8a5ec2b73d0915c6d95402d9ab064c5afac47f6df9e254d80caf39bbe3dR49-R50

It's because here the function returns something borrowed from `self`, but `self` is only borrowed for the lifetime of the function. So when that function returns a value that borrows from `self`, it is no longer valid.

`self` can be also borrowed for the lifetime declared on the struct, but that creates a lot of other problems.

So not sure how to solve this with references.

PR #393, opened 2025-06-09
fix(api): add missing resolver cache to a custom create app method

ilchu:
### Description

I broke the CI by not altering one of the new tests after rebase, so this is a one-line fix. 

### Changes

- Fixed method in `api/methods/block_number`


code-dogg (approved):
Oh no

PR #397, opened 2025-06-12
fix(server): Change `max_readers` value `20` => `16348`

code-dogg:
### Description
The readers are not being cleared after redeploy, so their number would eventually become permanently exhausted.

This is most likely caused by a lack of shutdown cleanup process is implemented.

Until this process is implemented, the number is increased.

### Changes
- Increase the number of max readers

### Testing
:green_circle: CI
:green_circle: Docker


birchmd (approved):
Could you also file an issue for the clean shutdown and mark this with a TODO?

PR #400, opened 2025-06-12
Fix(integration-test): EVM storage and missing transaction handling

birchmd:
### Description
This PR fixes the integration so that it works again (with all backends).

### Changes
- Failing to find a transaction by hash is not an error. This is important to get right because libraries for interacting with Ethereum RPCs handle null responses differently from errors and will check for transactions by hash in a loop when waiting for a status.
- Use the same EVM account storage for both application reader and writer for the in-memory backend. This is necessary because otherwise the reader does not see the changes the writer makes. The database backends already had this property because the reader and writer refer to the same DB on-disk.
- Use `Arc<RwLock<HashMap>>` for in-memory block storage. I don't know why the `evmap` implementation doesn't work, but the integration test fails to show op_node the right genesis block (from geth), so it must have something to do with the eventual consistency.
- Create the genesis block during app initialization. This is necessary because the app panics if the block storage is ever empty, which could happen if it receives a some other request before `eth_getBlockByNumber` (which is when we steal the genesis block from op-geth.

### Testing
- Integration test.

### DIFF
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -1,5 +1,3 @@
-#[cfg(test)]
-use umi_blockchain::block::{Block, BlockHash, ExtendedBlock, Header};
 use {
     crate::mirror::MirrorLog,
     clap::Parser,
@@ -16,7 +14,7 @@ use {
     umi_api::method_name::MethodName,
     umi_app::{Application, ApplicationReader, Command, CommandQueue, Dependencies},
     umi_blockchain::{
-        block::BlockQueries,
+        block::{Block, BlockHash, BlockQueries, ExtendedBlock, Header},
         payload::{NewPayloadId, StatePayloadId},
     },
     umi_genesis::config::GenesisConfig,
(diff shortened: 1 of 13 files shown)


server/src/lib.rs
```diff
@@ -166,11 +164,12 @@ pub fn initialize_app(
             &mut app.evm_storage,
         );
     }
+    let genesis_block = create_genesis_block(&app.block_hash, &genesis_config);
+    app.genesis_update(genesis_block);
```
code-dogg, line 168:
```suggestion
```

The genesis block is fetched from op-geth. Until we create our own genesis block, this should not be here, otherwise there will be two genesis blocks, one of them at index 1.

code-dogg (changes requested):
Neat fix!

server/src/lib.rs
birchmd: We need our own genesis block present because the server can get other queries before the one which steals the genesis block from op-geth. Without having our own genesis the node panics on those queries.

code-dogg: The first thing that op-node does is that it fetches genesis and all the block tags, which triggers the op_geth genesis stealing. Do we need to be worried about receiving queries before that? Maybe it's worse to respond to some queries with a different genesis than panic.

birchmd: In the integration test we call `eth_getBalance` from a separate process and its query can arrive before the first op_node query. I think this genesis inconsistency is a necessary evil until we can resolve #225

Issue #45, opened 2024-08-08
op-move config file


birchmd:
Options for running the `op-move` binary:

- ports to listen on
- directory for persistent storage
- etc.

Genesis config includes many fields, some of which are files. All the details of the config should be contained in a single file.

As we add more genesis details, this file can be extended.

Code references:
https://github.com/MovedNetwork/op-move/blob/5bff3017103056dbc9c1b2d4a300fc22089fd81e/genesis/src/config.rs#L25

https://github.com/MovedNetwork/op-move/blob/5065d42637a91b01a04a2c4608e547a58e71d266/server/src/lib.rs#L90

PR #402, opened 2025-06-14
feat(server): Add `umi-server-args` crate that declares parameters for running the server

code-dogg:
### Description
Adds a small crate that declares server settings.

### Changes
- Add `umi-server-args` crate

### Testing
:green_circle: CI

### Notes
#45

### DIFF
--- a/server/args/Cargo.toml
+++ b/server/args/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "umi-server-args"
+description = "Umi execution HTTP server settings"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+toml.workspace = true
+optional_struct.workspace = true
+serde.workspace = true
--- a/server/args/Cargo.toml
+++ b/server/args/Cargo.toml
@@ -6,5 +6,4 @@ edition = "2021"
 
 [dependencies]
 toml.workspace = true
-optional_struct.workspace = true
 serde.workspace = true
--- a/server/args/src/declaration.rs
+++ b/server/args/src/declaration.rs
@@ -1,18 +1,18 @@
 use {serde::Deserialize, std::net::SocketAddr};
 
-#[derive(Deserialize, PartialEq, Debug, Clone)]
+#[derive(PartialEq, Debug, Clone)]
 pub struct Config {
     pub auth: AuthSocket,
     pub http: HttpSocket,
 }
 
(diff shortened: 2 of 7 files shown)


server/args/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "umi-server-args"
+description = "Umi execution HTTP server settings"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+toml.workspace = true
```
bulenttastan, line 8: nit: alphabetical sorting

server/args/src/declaration.rs
```diff
@@ -0,0 +1,25 @@
...
+    #[optional_wrap]
+    pub auth: AuthSocket,
+    #[optional_rename(OptionalHttpSocket)]
+    #[optional_wrap]
+    pub http: HttpSocket,
+}
+
+#[optional_struct::optional_struct]
+#[derive(Deserialize, PartialEq, Debug, Clone)]
+pub struct HttpSocket {
```
bulenttastan, line 16: nit: can we keep this struct under `AuthSocket` for better ordering

server/args/Cargo.toml
```diff
@@ -0,0 +1,10 @@
...
+optional_struct.workspace = true
```
birchmd, line 9: I don't really like adding a crate dependency that is preventing only a few lines of boilerplate. Supply chain attacks can happen, especially with crates that are not widely used.

I would rather we explicitly have one type for the config taken from environment (whether that be disk, command line or env vars) where all the fields are optional, and a separate type for the "logical" config where none of the fields are optional. This crate provides this functionality, but ties the two types together via a proc macro whereas I think it might be useful to allow the two types to evolve independently since it may not always be true that the logical config fields correspond 1:1 with the environmental config fields.

server/args/src/declaration.rs
```diff
@@ -0,0 +1,35 @@
+use {serde::Deserialize, std::net::SocketAddr};
+
+#[derive(Deserialize, PartialEq, Debug, Clone)]
```
birchmd, line 3: The non-optional configs shouldn't need any serde traits right? They will always be constructed from the optional versions instead of from serialization.

birchmd (approved):

server/args/src/declaration.rs
code-dogg: Yes, that's sharp

PR #403, opened 2025-06-14
feat(server-args): Implement `ConfigBuilder` that builds config from a stack of `Layer`s

code-dogg:
### Description
Implements a feature that allows for layering config loaded from multiple sources and define their priority.

### Changes
- Add `ConfigBuilder` that builds config from multiple `Layer`s
- Add `Layer` that parses a config from one source

### Testing
:green_circle: CI

### Notes
#45


birchmd (approved):
Nice! I really like this design.

PR #414, opened 2025-06-17
Chore(CI): Run integration test nightly

### LINKED ISSUE #413: Run integration test nightly
Sometimes our changes cause the integration test to fail. We don't noticed it until we randomly run it.
The goal is to have it run once a day without disrupting the typical unit test runs.

bulenttastan:
### Description
Run just the integration test using the latest Docker images once a day, nightly at midnight Eastern summer time. Closes #413 

### Changes
- New GitHub workflow to run the integration test
- Use docker images for less redundancy and save logs for debugging
- Remove the chain setup from the test
- Remove heartbeat thread

### Testing
✓ [Final test run passes](https://github.com/UmiNetwork/op-move/actions/runs/16249595736/job/45877392774?pr=414)

### Notes
Used `docker compose` instead of `docker stacks`. Running the containers as stacks hung the process for hours and `op-move` couldn't start properly, even locally.


.github/workflows/integration.yaml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions-rs/toolchain@v1
+        with:
+          default: true
+          toolchain: stable
+      - name: Free up disk space
+        run: rm -rf /opt/hostedtoolcache
+      - name: Test
+        run: cargo test test_on_ethereum
```
birchmd, line 24:
```suggestion
        run: cargo test --release test_on_ethereum
```

I think it might make sense to compile in release mode since `op-move` can be a little sluggish in debug mode.

birchmd (approved):

ilchu (changes requested):
Seems like it'll need all the foundry / geth / op stack setup in the runner to actually work properly. That will be quite a hassle 😞

birchmd (approved):
Great job! Having the integration test be simpler to set up as well as working in CI are both welcome improvements.

PR #421, opened 2025-06-18
Fix(test): Use median of 100 trials to check solidity_abi gas costs

birchmd:
### Description
No changes to production code. Fixes flaky test `solidity_abi::tests::test_gas_costs`.

### Changes
- Benchmark function with 100 trials and use the median duration as the metric for determining test pass/failure.

### Testing
`solidity_abi::tests::test_gas_costs`

### Notes
This change doesn't completely eliminate the chance of flakiness. If we assume all 100 trials are independent (perhaps not a good assumption since timing might depend on ambient system load, which could be highly correlated from moment to moment) then using the median of 100 trials essentially amplifies the pass/fail signal, as shown in the plot below. Essentially, if a single trial has a 60% chance of passing or better then median will pass close to 100% of the time, but not literally 100% of the time.
image

### DIFF
--- a/evm-ext/src/solidity_abi.rs
+++ b/evm-ext/src/solidity_abi.rs
@@ -478,7 +478,6 @@ mod tests {
     };
 
     #[test]
-    #[ignore = "flaky"]
     fn test_gas_costs() {
         let gas_params = NativeGasParameters::initial();
         let mut rng = rand::thread_rng();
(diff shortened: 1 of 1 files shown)


code-dogg (changes requested):
Hmm... How about we promote this test to a benchmark?

evm-ext/src/solidity_abi.rs
```diff
@@ -478,7 +478,6 @@ mod tests {
     };
 
     #[test]
-    #[ignore = "flaky"]
     fn test_gas_costs() {
```
code-dogg, line 481: Yes, it still fails sometimes. I don't think we're there yet.

```suggestion
    #[ignore = "flaky"]
    fn test_gas_costs() {
```

bulenttastan: Good catch.
What's expected here, that it passes 65% of the time?

birchmd: I have created #425 to address this long term. Leaving the test marked as flaky for now.

code-dogg (approved):

PR #423, opened 2025-06-18
Feat(transactions): Include table changes

### LINKED ISSUE #384: Include changes from tables extension
We have [a place to capture tables changes](https://github.com/UmiNetwork/op-move/blob/4ab2338109ae16d61fffd78d96b35634e8a7d11f/state/src/diff.rs#L20), but it is never populated from running a transaction. This needs to be fixed for the tables to be updated in the state. The changes will need to be extracted from the extensions context. We must also charge gas for the storage costs of these state changes.

birchmd:
### Description
Fixes #384 

### Changes
- Capture changes from `NativeTableContext` and include them in the changes returned by executing canonical transactions.
- Charge gas for the new changes.

### Testing
- Update `test_execute_tables_contract` to use the normal transaction execution flow instead of custom handling of tables.

### DIFF
--- a/state/src/lib.rs
+++ b/state/src/lib.rs
@@ -18,7 +18,6 @@ use {
     move_binary_format::errors::{Location, VMResult},
     move_core_types::{
         account_address::AccountAddress,
-        effects::ChangeSet,
         identifier::IdentStr,
         language_storage::{ModuleId, StructTag},
     },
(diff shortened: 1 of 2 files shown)


birchmd:
> table changes still need to be included in the trie

Good catch @code-dogg ! Done in https://github.com/UmiNetwork/op-move/pull/423/commits/970c0de303f8b04b9aba7547ebdae3ef400ab005

PR #424, opened 2025-06-19
refactor(txs): normalize txs earlier

### LINKED ISSUE #365: Reject unsupported transaction types
As mentioned in https://github.com/UmiNetwork/op-move/pull/362#discussion_r2104643171, we are currently not returning any errors if a transaction is submitted to a mempool with an unsupported type, such as EIP-7702 or EIP-4844. We should be returning an error in such cases instead of panicking or ignoring as we do now.

ilchu:
### Description

This PR enforces normalization of transactions right upon receival, which has a benefit of
doing less unwraps / panics within execution flow due to bubbling up of unsupported variants or failed signer recovery. This also makes our transaction flow more explicit, gathering all the conversions and allowed variants in one place. 

Closes #365.

### Changes

- Removes `MempoolTransaction` and fully relies on the `NormalizedEthTransaction` instead
- Introduces some conversion between our envelopes and OpTxEnvelope
- Expands the normalized envelope to also hold L1 gas inputs, tx hashes and signatures.

### Testing

- [x] Old tests passing

### Notes

~I also plan to get rid more thoroughly of the `OpTxEnvelope` usage within this PR, like how it's used in `ExecutionPayload` txs. doing this cleanup I'll undraft this PR.~

### DIFF
--- a/app/src/input.rs
+++ b/app/src/input.rs
@@ -37,7 +37,7 @@ impl TryFrom<Payload> for PayloadForExecution {
     type Error = umi_shared::error::Error;
 
     fn try_from(value: Payload) -> Result<Self, Self::Error> {
-        let mut transactions = Vec::new();
+        let mut transactions = Vec::with_capacity(value.transactions.len());
 
         for raw_tx in value.transactions {
             let mut slice: &[u8] = raw_tx.as_ref();


storage/rocksdb/src/transaction.rs
```diff
@@ -117,16 +115,17 @@ mod tests {
...
         let transaction = ExtendedTransaction {
-            inner: signed_tx,
+            inner: normalized_tx.into(),
             block_number: 1,
             block_hash: B256::new(hex!(
                 "2222223123123121231231231231232222222231231231212312312312312322"
             )),
             transaction_index: 1,
-            effective_gas_price: 1,
+            effective_gas_price: U256::from(1),
```
birchmd, line 128: Nit: `U256` has a constant for 1 we could use instead of the conversion function.

execution/src/transaction.rs
```diff
@@ -25,45 +29,131 @@ pub const L2_LOWEST_ADDRESS: Address = address!("4200000000000000000000000000000
...
-                    logs_bloom: bloom,
-                })
-            }
-            _ => unreachable!("Wrapping of other tx types not supported"),
         }
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Deserialize, Serialize)]
```
birchmd, line 136: This comment applies everywhere serde traits are newly added in this PR.

I get nervous when we add serde traits onto data types we use for internal business logic. It means we can break backwards compatibility (making upgrades challenging) accidentally. For example, if we realize a different representation of the type makes some processing more efficient we have to worry about if the change will cause incompatible serialization.

This is why I generally recommend having completely separate types for serialization (i.e. external IO) and internal use.

What change in this PR causes us to require serialization logic on this type? Could we keep thigs so that the op-alloy type is still used at the IO boundaries and convert to our internal type before passing it into the transaction processing flow (including the mempool)?

execution/src/transaction.rs
```diff
@@ -25,45 +29,131 @@ pub const L2_LOWEST_ADDRESS: Address = address!("4200000000000000000000000000000
...
 pub enum NormalizedExtendedTxEnvelope {
     Canonical(NormalizedEthTransaction),
-    DepositedTx(TxDeposit),
+    DepositedTx(Sealed<TxDeposit>),
+}
+
+// TODO: avoid conversion to op-alloy and do it internally
+impl Decodable for NormalizedExtendedTxEnvelope {
```
birchmd, line 143: - [ ] Similarly with the serde traits, I wonder if this impl can be avoided by keeping op-alloy types at all IO boundaries.

execution/src/transaction.rs
```diff
@@ -109,6 +212,44 @@ impl NormalizedExtendedTxEnvelope {
...
+        match self {
+            Self::Canonical(tx) => tx.l1_gas_fee_input.clone(),
+            Self::DepositedTx(_) => L1GasFeeInput::default(),
+        }
+    }
+
+    pub fn wrap_receipt(&self, receipt: Receipt, bloom: Bloom) -> OpReceiptEnvelope {
+        match self {
+            Self::Canonical(_) => {
+                // For canonical transactions, we use EIP-1559 receipt format
```
birchmd, line 233: Shouldn't the receipt type depend on the transaction type?

execution/src/transaction.rs
```diff
@@ -199,30 +344,22 @@ impl NormalizedEthTransaction {
...
 impl TryFrom<Signed<TxEip1559>> for NormalizedEthTransaction {
     type Error = Error;
 
     fn try_from(value: Signed<TxEip1559>) -> Result<Self, Self::Error> {
         let address = value.recover_signer()?;
-        let tx = value.strip_signature();
+
+        let mut encoded = Vec::new();
+        value.rlp_encode(&mut encoded);
+        let l1_gas_fee_input = encoded.as_slice().into();
```
birchmd, line 355: I don't really like needing to completely re-encode the transaction to get this information. Can the flow be modified somehow so that instead of using conversion traits we have constructor methods that take in the gas information as an argument?

birchmd (commented):
Nice work on this so far. I like the direction this is going in terms of pushing the tx validation boundary all the way back to the source of transactions.

My only concern is with adding serde traits on data types we use in business logic. Can we refactor things so that they are not needed?

execution/src/transaction.rs
ilchu: Made it into a builder-like method in e6e5b3e.

execution/src/transaction.rs
ilchu: This was due to the [change](https://github.com/UmiNetwork/op-move/pull/424/files#diff-c1e103fa3ecb989bb4c6ffacbbeb5f72ace10eeeacd5fc87db159fcba8c980f4L10-R12) of the inner type in `ExtendedTransaction`. I reverted it to using `OpTxEnvelope` internally, but still forcing it to be constructed from a normalized envelope to avoid side channel conversions in e459329.

blockchain/src/transaction/write.rs
```diff
@@ -1,59 +1,45 @@
...
     std::fmt::Debug,
+    umi_execution::transaction::NormalizedExtendedTxEnvelope,
     umi_shared::primitives::{B256, U256},
 };
 
-#[derive(Debug, Clone, Hash, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, Hash, Eq, PartialEq, Serialize, Deserialize)]
 pub struct ExtendedTransaction {
     pub inner: OpTxEnvelope,
+    pub from: Address,
```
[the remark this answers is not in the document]
ilchu: Indeed, we have to be careful about these. But also all the more reason to be active about these changes while we have the freedom to do it without further complications.

PR #426, opened 2025-06-19
fix: Remove static database references

code-dogg:
### Description
Removes static database variables and makes them instance-based instead.

This makes the `Drop` implementation work, which may do some cleanup.

For example `heed`: https://docs.rs/heed/latest/heed/struct.Env.html#method.prepare_for_closing

Static variables are never dropped and that causes the LMDB do accumulate stale readers on every deploy. Having a stale reader means that there is a reader that no longer exists holding a spot for a reader, whose number is limited. The maximum number of readers is configurable, but without having them cleaned up there cannot be a way to set it to a number that's based on some real limitations.

### Changes
- Remove all `static` variables

### Testing
Running `./docker/host/deploy.sh` repeatedly while setting a low number of `max_readers` in the `heed` configuration in `create_db` function.

### Notes
A bunch of `Arc`s are added in this. In the case of `heed`, the `Arc` is actually on the [inside](https://github.com/meilisearch/heed/blob/main/heed/src/envs/env.rs#L32).

I think the `Arc`s can be avoided by sharing one reference instead. But that reference needs to be borrowed from one layer above `Dependencies`, i.e. the `Dependencies` cannot actually own it (and therefore cannot create the database) because the lifetime of `Dependencies` is much shorter than the lifetime of the database.

Well, I decided to not bother with this for now.

### DIFF
--- a/server/src/dependency/heed.rs
+++ b/server/src/dependency/heed.rs
@@ -28,14 +28,14 @@ impl HeedDependencies {
     }
 }
 
-impl umi_app::Dependencies for HeedDependencies {
+impl<'db> umi_app::Dependencies<'db> for HeedDependencies {
+    type BlockHashLookup = SharedBlockHashCache;
+    type BlockHashWriter = SharedBlockHashCache;
     type BlockQueries = block::HeedBlockQueries<'static>;
     type BlockRepository = block::HeedBlockRepository<'static>;
-    type BlockHashLookup = umi_app::SharedBlockHashCache;
-    type BlockHashWriter = umi_app::SharedBlockHashCache;
-    type OnPayload = umi_app::OnPayload<Application<Self>>;
-    type OnTx = umi_app::OnTx<Application<Self>>;
-    type OnTxBatch = umi_app::OnTxBatch<Application<Self>>;
+    type OnPayload = umi_app::OnPayload<Application<'db, Self>>;
+    type OnTx = umi_app::OnTx<Application<'db, Self>>;
+    type OnTxBatch = umi_app::OnTxBatch<Application<'db, Self>>;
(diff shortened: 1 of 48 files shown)


server/src/dependency/heed.rs
```diff
@@ -290,7 +271,7 @@ fn create_db() -> umi_storage_heed::Env {
 
     let env = unsafe {
         EnvOpenOptions::new()
-            .max_readers(16348) // TODO: Cleanup readers on shutdown
```
birchmd, line 293: Is the TODO still needed?

server/src/dependency/in_memory.rs
```diff
@@ -13,6 +13,8 @@ pub fn dependencies() -> Dependency {
     InMemoryDependencies::new()
 }
 
+unsafe impl Sync for InMemoryDependencies {}
```
birchmd, line 16: I know that we talked about the unsafe impls before, but I forget if we have an issue for it or not. Please add a TODO and issue number for reviewing (and eliminating) unsafe code. I don't think there is anything we are doing that should require it.

It's lower priority of course since the in-memory implementation is only used in tests.

birchmd (approved):
Looks good to me.

server/src/dependency/heed.rs
code-dogg: No that should not be needed, the readers are cleared using the `Drop` implementation.

code-dogg: Will do that separately

server/src/dependency/in_memory.rs
code-dogg: #427

Didn't add the TODO comment though, merged too fast. Maybe we can add it separately.

PR #428, opened 2025-06-22
feat(server-args): Add database configuration

code-dogg:
### Description
Adds database parameters into the main `Config`.

### Changes
- Add `Database` into `Config`
- Use the `Database` parameters in `create_db` functions

### Testing
:green_circle: Fmt
:green_circle: Clippy
:green_circle: Test

### Notes
#45

### DIFF
--- a/server/args/src/declaration.rs
+++ b/server/args/src/declaration.rs
@@ -1,7 +1,7 @@
 use {
-    clap::{Args, Parser},
+    clap::{Args, Parser, ValueEnum},
     serde::Deserialize,
-    std::net::SocketAddr,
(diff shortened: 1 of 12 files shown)


server/args/src/declaration.rs
```diff
@@ -23,12 +24,32 @@ pub struct HttpSocket {
     pub addr: SocketAddr,
 }
 
+#[derive(PartialEq, Debug, Clone)]
+pub struct Database {
+    /// TODO: Currently a dummy, either make it work or remove it
```
birchmd, line 29: I think it is worth investing the time to make this config option work.

birchmd (approved):

PR #431, opened 2025-06-24
Feat: keep track of in-progress payloads to avoid unknown payload errors

birchmd:
### Description
Adds a small in-memory cache for in-progress forkchoice update payloads (i.e. blocks). This cache allows the RPC to delay a bit for a block to finish building before returning a result from `engine_getPayload`, this preventing some unknown payload errors. The purpose of this change is to try to address the slowness of the faucet on Devnet.

### Changes
- New synchronous (via `RwLock`) in-memory cache of in-progress blocks.
- RPC uses this cache to know when it should wait for a block to finish processing instead of returning an unknown payload error.

### Testing
- Integration testing shows this change does have an improvement on deposit times

### Notes
- This change will not necessarily stop all unknown payload errors. If the `engine_getPayload` call comes before the block even starts building (e.g. because the command queue is very backed up) then this in-progress tracker will not help. However, I think this change should prevent errors most of the time because usually `op-move` will start building the block right away and thus it will be in progress before the `engine_getPayload` RPC call arrives.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12349,6 +12349,7 @@ dependencies = [
  "serde",
  "sha2 0.10.9",
  "test-case",
+ "tokio",
  "umi-evm-ext",
  "umi-execution",
  "umi-genesis",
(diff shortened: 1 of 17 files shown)


code-dogg (approved):
This fixes a problem and took some time and effort so I have no problem having this merged. I have some design questions, but I feel like it's too late to question design decisions during a code review - those should be made and addressed before the code is written.

It mainly concerns this. The responsibility of keeping track of in-progress payloads is moved from the client 2 layers deeper into the application. It introduces some grey area between the `queue.send` and the `in_progress_payloads.start_on`. How would it look like if it was solved at the API layer?

PR #433, opened 2025-06-25
ci: Add `concurrency` option into `build` workflow to cancel old runs

code-dogg:
### Description

Insert description here

### Changes

- Change 1
- Change 2


code-dogg:
Well the workers are in the cloud and every workflow starts from a fresh ubuntu docker image so no cross contamination or taking up space is happening cross workflows.

Also, since this is a public repository, there is no limit on free CI minutes as is the case for private repositories.

.github/workflows/build.yaml
```diff
@@ -6,6 +6,10 @@ on:
       - main
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
birchmd, line 10: How does the `github.event.pull_request.number` resolve in the case that CI triggers on a push to `main`?

code-dogg: Defaults to `github.ref`

birchmd: Ah, ok. And `github.ref` is the commit hash or something?

PR #434, opened 2025-06-26
fix(server): pass manifest dir to config default layer

ilchu:
### Description

I noticed there's a difference between how genesis config is treated in tests vs production env. And this caused the integration tests to fail with file not found on the token list, so this simple fix should do the trick.

### Changes

- Do the same thing in regard to file paths as in the `Default` impl of the `GenesisConfig`

### Testing

Integration passes again locally.


code-dogg (changes requested):
I think the paths need to be adjusted here as well.

https://github.com/UmiNetwork/op-move/blob/f945f3a5025da5bb17e52fdd531926c106e4e682/docker/Dockerfile.op-move#L55

PR #440, opened 2025-06-27
build(docker): Add cache configuration

code-dogg:
### Description
Adds cache configuration into `deploy` workflow.

In the best case it shaves off over an **hour and a quarter** as you can see in the difference between the two attempts of this workflow run https://github.com/UmiNetwork/op-move/actions/runs/15927637949.

### Changes
- Add `cache_from` and `cache_to` imports and exports of cache layers

### Testing
Using the CI


ilchu (approved):
One and a half hour is wild! Great job

birchmd (approved):

PR #443, opened 2025-06-30
Fix(blockhash-cache): Use cache properly in DB backends

birchmd:
### Description
`test_blockhash_evm_contract` was failing on the DB-based storage backends because the cache was not being used properly there. This PR fixes this issue and removes some unnecessary code. Note: this test failing means the `BLOCKHASH` opcode is currently broken on devnet.

### Changes
- Make `HybridBlockHashCache` shared, delete `SharedHybridBlockHashCache`.
- Use `HybridBlockHashCache` for all `BlockHashLookup`/`BlockHashWriter` implementations.
- Clone the same `HybridBlockHashCache` instance instead of creating a fresh one in the DB-based storage backends.

### Testing
- `test_blockhash_evm_contract`


api/src/methods/mod.rs
```diff
@@ -247,28 +248,28 @@ pub mod tests {
...
+                    (),
+                    (),
+                    (),
+                    (),
+                    (),
+                    (),
+                    (),
+                    (),
+                    HybridBlockHashCache<(), StubLatest>,
+                    HybridBlockHashCache<(), StubLatest>,
```
ilchu, line 270: I wish this was less painful. Honestly I couldn't figure out why rustc failed to infer some type parameters and was perfectly fine with some others, so it was a guessing game when writing this.

Also as a side question, do you feel we would benefit from some more systematic mocking? We have this test dependencies instantiation in a few different places and it could be more nicely centralized.

app/src/block_hash.rs
ilchu, line 1: Happy to see my accidental complexity gone 😅

server/src/dependency/heed.rs
```diff
@@ -63,11 +68,11 @@ impl<'db> umi_app::Dependencies<'db> for HeedDependencies {
     type TransactionRepository = transaction::HeedTransactionRepository<'db>;
 
     fn block_hash_lookup(&self) -> Self::BlockHashLookup {
-        SharedBlockHashCache::new()
+        self.block_hash_lookup.clone()
```
ilchu, line 71: So the crux of this bug was that a single block hash cache instance should have been shared between the lookup and reader, but they were two separate unsynchronized ones, I suppose.

ilchu (approved):
Awesome job catching this. One more call for us to consider running tests with proper storage flags in CI too.

api/src/methods/mod.rs
birchmd: Yes, I completely agree. I had a hard time filling these in too. I think the core issue is that Rust's syntax isn't really meant to scale to this number of type generics. I'm not sure if there is a good way to make this more developer friendly.

Centralizing the testing implementations would be a welcome change! Probably lower priority than other things though.

code-dogg (approved):
Good job reducing the code complexity

PR #444, opened 2025-06-30
Chore(integration-test): Run op-move on the same runtime as the test runner

birchmd:
### Description
I find having `op-move` executing in a separate Runtime can obscure some errors because tokio will complain about not being able to drop a Runtime from inside an async function, but of course the reason it is trying to drop a Runtime in an async function in the first place is because something else has gone wrong. This PR changes the integration test setup so that `op-move` is spawned as a task in the existing Runtime instead of creating a new one. The integration test still passes for me, so I think this is a safe change to make and might make debugging easier in case it is failing in the future.

### Changes
- No change to production code. 

### Testing
- Integration test

### DIFF
--- a/server/src/tests/integration.rs
+++ b/server/src/tests/integration.rs
@@ -27,7 +27,7 @@ use {
         str::FromStr,
         time::{Duration, Instant},
     },
-    tokio::{fs, runtime::Runtime},
+    tokio::fs,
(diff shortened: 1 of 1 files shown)


code-dogg (approved):
Yeah that was just obfuscating the real error.

PR #448, opened 2025-07-02
Fix(blockhash-cache): initialize on startup

### LINKED ISSUE #419: Fix block hash cache initialization
For reference, see #418

To avoid accessing database for `ApplicationReader` dependency initialization, currently the cache is initialized empty.

However, it seems that the cache should be filled for it to work properly.

The goal of this issue is to fill the cache in, or implement some other way to make it work, but not by making it a condition for initialization.

birchmd:
### Description
Closes #419 

### Changes
- Try to initialize the blockhash cache on startup for DB-based backends; fallback on a new cache in case of failure.

### Testing
@code-dogg would you mind testing if this works with the docker setup?


bulenttastan (approved):
Looks good to me, verified with a Solidity file.

app/src/block_hash.rs
```diff
@@ -125,17 +118,36 @@ where
...
+    }
+
+    fn initialize_then<Out, F>(&self, f: F) -> Out
+    where
+        F: FnOnce(&mut BlockHashRingBuffer) -> Out,
+    {
+        let mut maybe_buffer = self.ring_buffer.write().expect("Lock not poisoned");
+        let buffer = maybe_buffer.get_or_insert_with(|| {
+            BlockHashRingBuffer::try_from_storage(&self.storage, &self.block_query)
+                .unwrap_or_default()
```
code-dogg, line 140: Is a default value on failure acceptable? Wouldn't it cause the same issue as with uninitialized blockhash cache?

birchmd: Yes, the default value is the same as if the cache is not initialized, however the cache is purely an optimization because the `hash_by_number` function always falls back on the `block_query` if the number is not found in the cache. So the cache is not necessary for correctness.

That said, `block_query` is necessary for correctness and that is also what is used to build up the cache in `try_from_storage`. So if the cache is not successfully populated then the fallback to the cache will also not work properly.

It is reasonable to propagate the error back up to the EVM native where it can cause execution to terminate, but I think that should be done as a separate PR.

code-dogg (approved):

Issue #86, opened 2024-08-29
Transform ERC20 calls internally to Move function calls


bulenttastan:
Getting an Ethereum call would send RLP bytecode instead of BCS encoded calls. This would be considered an incorrect call to run a Move entry function. However we should make an exception and handle RLP calls made specifically for ERC20 and ERC721 methods so that we can for instance enable token transfers from MetaMask.

The ERC20 and ERC721 methods are standard, so we should be able to figure out that a user is making these calls and map them to appropriate Move entry functions on the corresponding token modules.

This is not an issue for ETH transfers because that relies on the `value` parameter and that's already being handled.

bulenttastan:
@birchmd should we separate this issue to work on it independent of the ERC20 design?

bulenttastan:
We have a team working on an NFT project. We can prioritize this work, at least the ERC721 case.
The goal is to provide the expected response so that MetaMask can display the NFTs from a Move contract.

PR #449, opened 2025-07-02
Feat(transactions): Automatically recognize ERC20 transactions

birchmd:
### Description
This PR does part of the work for #86 (only dealing with ERC-20 tokens). The purpose of adding this functionality is to allow Ethereum wallets like Metamask to interact with bridged tokens on Umi exactly the same way as they would on EVM-based L2s. Since these wallets would send transactions containing Solidity ABI-encoded data directly (without any of our custom wrappings to distinguish it from BCS-encoded transaction data), our node must be able to correctly interpret these transactions.

Note that this change does not automatically map an ERC-20 method into any kind of Move-based token method, it only delegates to the EVM. This is intentional because the interface for Move-based tokens might be different. It is recommended that dApp developers using Move-based tokens also build an appropriate front-end for their users to interact with their token balances. However, tokens bridged from Ethereum really are ERC-20 tokens, so this mechanism will work well with them.

### Changes
- Recognize when transaction data encodes an ERC-20 method and automatically interpret it as an EVM call.

### Testing
- Tests in `execution/src/tests/erc20` are generalized so they try to interact with ERC-20 tokens both with transactions encoding Move types and those using the Solidity ABI directly.

### Notes
- The bulk of this PR is changes to the ERC-20 tests. To isolate the functional changes for production just look at changes from the first commit.


execution/src/transaction.rs
```diff
@@ -508,7 +508,16 @@ impl TransactionData {
                     // transaction as a base token transfer between EOAs.
                     Ok(Self::EoaBaseTokenTransfer(to))
                 } else {
-                    let tx_data: SerializableTransactionData = bcs::from_bytes(&tx.data)?;
+                    let tx_data: SerializableTransactionData =
+                        bcs::from_bytes(&tx.data).or_else(|e| {
```
bulenttastan, line 512: I think we should leave a short comment saying if it's not BCS deserializable it could be a direct ERC20 contract interaction.

execution/src/tests/erc20/move_impl.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn decimals(ctx: &TestContext, token_address: Address) -> u8 {
+        let outcome = ctx
+            .quick_call(
+                vec![MoveValue::Address(token_address.to_move_address())],
+                "erc20",
+                "decimals",
+            )
+            .0;
+        let val = U256::from_be_slice(&outcome.output);
+        val.as_limbs()[0] as u8
```
bulenttastan, line 120: Does this need a check for other limbs to be zero?

bulenttastan (approved):
It's slightly flaky but it mostly works. Here's a screenshot on Metamask.

Screenshot 2025-07-03 at 1 41 04 AM

Issue #203, opened 2025-01-24
Audit error handling of `unwraps`, client issues and where user errors shouldn't panic


bulenttastan:
This is to verify there aren't unnecessary `unwrap`s in the codebase that could bring down the server. Client issues and user errors should also not panic, which would shut down the server.

PR #453, opened 2025-07-09
Feat(stability): remove/replace unwraps

birchmd:
### Description
Works towards #203 by checking all `unwrap` usages. Most are replaced with error propogation, some are replaced by `expect` calls with a description of why it should never panic. Any new `unwrap`s in the diff occur in tests where a function now returns a `Result` where is did  not before.

### Changes
- Remove/replace `unwrap` calls (refactor only).

### Testing
- Existing tests.

### Notes
- Usages of `unwrap` found via `cargo clippy -- -Dclippy::unwrap_used`
- Reviewing this PR may be easier if done commit by commit instead of all at once.

### DIFF
--- a/app/src/actor.rs
+++ b/app/src/actor.rs
@@ -24,10 +24,17 @@ pub struct CommandActor<'actor, 'app, D: Dependencies<'app>> {
     app: &'actor mut Application<'app, D>,
 }
 
+// Marker struct for errors where the app has no choice but to
+// stop receiving commands and drop.
+#[derive(Debug)]
+pub(crate) struct UnrecoverableAppFailure;
+
 impl<'app, D: DependenciesThreadSafe<'app>> CommandActor<'_, 'app, D> {
     pub async fn work(mut self) {
         while let Some(msg) = self.rx.recv().await {
-            Self::handle_command(&mut *self.app, msg);
+            if Self::handle_command(&mut *self.app, msg).is_err() {
+                break;
+            }
         }
     }
 }
@@ -37,14 +44,19 @@ impl<'actor, 'app, D: Dependencies<'app>> CommandActor<'actor, 'app, D> {
         Self { rx, app }
     }
 
-    pub fn handle_command(mut app: impl DerefMut<Target = Application<'app, D>>, msg: Command) {
+    fn handle_command(
+        mut app: impl DerefMut<Target = Application<'app, D>>,
+        msg: Command,
+    ) -> Result<(), UnrecoverableAppFailure> {
         match msg {
             Command::StartBlockBuild {
                 payload_attributes,
                 payload_id,
             } => app.start_block_build(payload_attributes, payload_id),
-            Command::AddTransaction { tx } => app.add_transaction(tx),
-            Command::GenesisUpdate { block } => app.genesis_update(block),
+            Command::AddTransaction { tx } => {
+                app.add_transaction(tx);
+                Ok(())
+            }
         }
     }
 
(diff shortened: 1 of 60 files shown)


app/src/command.rs
```diff
@@ -48,15 +58,27 @@ impl<'app, D: Dependencies<'app>> Application<'app, D> {
...
         let parent = self
             .block_repository
             .latest(&self.storage)
-            .unwrap()
-            .expect("Parent block should exist");
+            .map_err(|e| {
+                tracing::error!("Failure during `start_block_build`. Failed to get latest block from block repository: {e:?}");
+                UnrecoverableAppFailure
+            })?
+            .expect("Block repository is non-empty (must always at least contain genesis)");
```
bulenttastan, line 81: Is the first transaction always the L1 deposit transaction instead of genesis? By genesis did you mean that?

bulenttastan (approved):
Left some comments, but otherwise looks good to me

app/src/command.rs
```diff
@@ -28,18 +29,27 @@ use {
...
             .payload_queries
             .by_id(&self.storage_reader, id)
-            .unwrap()
-            .is_some()
-        {
-            return;
+            .map_err(|e| {
+                tracing::error!(
+                    "Failure during `start_block_build`. Payload queries failed: {e:?}"
+                );
```
ilchu, line 43: Shouldn't the above `#[instrument]` macro 
already take care of the scoping? AFAICT it names the span identically as the function within which an event was created.

genesis-image/build.rs
```diff
@@ -19,6 +19,8 @@ fn main() {
     save(&vm, &genesis_config, &storage_trie);
 }
 
+// Safety: genesis-image is only used in tests
```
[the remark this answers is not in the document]
ilchu: Maybe the most precise way to say it would be that we prebuild the genesis for tests only, and in other cases always resort to doing it from scratch.

genesis-image/build.rs
```diff
@@ -19,6 +19,8 @@ fn main() {
...
+#[allow(clippy::unwrap_used)]
```
ilchu, line 23: On a second note, isn't this lint allowed by default? So we should explicitly deny it in `clippy.toml`. Also apparently there's a separate [setting](https://doc.rust-lang.org/clippy/lint_configuration.html#allow-unwrap-in-tests) for allowing it in tests automatically.

ilchu (commented):
Always great to see our system getting more robust. Just a few minor comments.

app/src/actor.rs
```diff
@@ -37,14 +44,19 @@ impl<'actor, 'app, D: Dependencies<'app>> CommandActor<'actor, 'app, D> {
...
+        mut app: impl DerefMut<Target = Application<'app, D>>,
+        msg: Command,
+    ) -> Result<(), UnrecoverableAppFailure> {
         match msg {
             Command::StartBlockBuild {
                 payload_attributes,
                 payload_id,
             } => app.start_block_build(payload_attributes, payload_id),
-            Command::AddTransaction { tx } => app.add_transaction(tx),
-            Command::GenesisUpdate { block } => app.genesis_update(block),
```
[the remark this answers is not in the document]
birchmd: Yes, the `GenesisUpdate` command is no longer used, so I removed it.

genesis-image/build.rs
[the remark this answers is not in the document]
birchmd: By `genesis-image` I mean this whole crate. The point is that unwraps here are fine because it isn't production code.

genesis-image/build.rs
birchmd: Yes, this lint is allowed by default. I was running clippy with `-Dclippy::unwrap_used` CLI flag to find the `unwraps` to audit. I haven't added a CI job to check for unwraps, but we could do that now.

PR #455, opened 2025-07-10
Fix(stability): Audit usage of `expect`

### LINKED ISSUE #203: Audit error handling of `unwraps`, client issues and where user errors shouldn't panic

birchmd:
### Description
Closes #203 
I audited all the `expect` calls in the code base. Most of them are about serialization (which should always succeed) or setting up genesis (which should cause panic if it fails). There were some that were part of unused code paths, so I simply deleted that unused code. In one case the `expect` could be replaced with `?`.

### Changes
- Remove unused `FromOptRoot` trait
- Remove unused proxy to op-geth
- Replace expects with `?` in fee_history implementation

### Testing
- Existing tests

### DIFF
--- a/app/src/query.rs
+++ b/app/src/query.rs
@@ -374,12 +374,11 @@ impl<'app, D: Dependencies<'app>> ApplicationReader<'app, D> {
             .hashes()
             .map(|hash| {
                 let rx = self
-                    .transaction_receipt(hash)
-                    .expect("Database should work")
-                    .expect("Tx receipt should exist");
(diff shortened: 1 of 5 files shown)


bulenttastan (approved):
✓ Integration test passes

ilchu (approved):
LGTM

PR #456, opened 2025-07-12
feat(api): `eth_gasPrice`, `eth_maxPriorityFeePerGas` endpoints

### LINKED ISSUE #267: Return actual gas price instead of hard coded value
Gas price value is currently not exposed correctly, but it was needed for SDK access. So we provided a hard coded value. We should give the correct value which changes depending on congestion.

https://github.com/MovedNetwork/op-move/blob/67cb13aa89f23704019129285f6d20d8ed8d6823/api/src/methods/gas_price.rs#L13

ilchu:
### Description

Adds actual logic for `eth_gasPrice` instead of a placeholder and introduces another gas fee prediction endpoint that's part of the current Ethereum RPC spec.

Closes #267.

### Changes

- Unified 0 parameter handling in API crate
- The actual gas price algorithm in `app/src/query.rs`
- A new endpoint for `eth_maxPriorityFeePerGas` method that's not legacy and is functionally same, except the non-addition of base fee to the end result
- New tests

### Testing

- [x] new API tests
- [x] new tests in `app/src/tests.rs`

### Notes

Following `op-geth` example, we chose the newer simplified algorithm introduced by Optimism, that assumes a single builder, instead of the default `geth` one that's relying on gas price oracles and is closer in complexity to fee history calculation.

### DIFF
--- a/api/src/method_name.rs
+++ b/api/src/method_name.rs
@@ -19,6 +19,7 @@ pub enum MethodName {
     TransactionReceipt,
     GetProof,
     GasPrice,
+    MaxPriorityFeePerGas,
 }
 
 impl MethodName {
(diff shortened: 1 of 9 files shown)


bulenttastan (commented):
Looks good to me other than 1 comment. I'll leave others to review too.

Issue #415, opened 2025-06-17
API to get contract data from both EVM and MoveVM


bulenttastan:
Implement `eth_getCode` for EVM to get contract bytecode.

MoveVM has more APIs to access the data:[module / resource](https://github.com/aptos-labs/aptos-core/blob/main/api/src/state.rs)  and [table data](https://github.com/aptos-labs/aptos-core/blob/main/api/src/state.rs) with handle id.

PR #464, opened 2025-07-20
feat(blockchain): Add `move_bytecode_at` state query for reading Move module bytecode

code-dogg:
### Description
Adds a query for reading Move module bytecode for an arbitrary height.

### Changes
- Add `move_bytecode_at` state query for reading Move module bytecode

### Testing
Using clippy check

### Notes
#415

### DIFF
--- a/blockchain/src/state/read/model.rs
+++ b/blockchain/src/state/read/model.rs
@@ -6,7 +6,10 @@ use {
     },
     eth_trie::{DB, EthTrie, Trie},
     move_binary_format::errors::PartialVMError,
-    move_core_types::{account_address::AccountAddress, vm_status::StatusCode},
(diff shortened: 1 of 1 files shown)


blockchain/src/state/read/model.rs
```diff
@@ -143,6 +146,22 @@ pub trait StateQueries {
         Ok(Some(bytes.into()))
     }
 
+    fn move_bytecode_at(
+        &self,
+        account: AccountAddress,
+        name: &str,
```
bulenttastan, line 152: This code looks good. How can we implement the `listing` function to get all the modules for an account? I'm curious, cause that might affect this method.

birchmd (approved):

PR #466, opened 2025-07-23
Fix(eth_call): include block parameters

### LINKED ISSUE #465: Fix eth_call to return global block data
Block details (ie blockhash, block number) are exposed as global data through `sendRawTransaction` but not from the simulation call.

Also the chain id is set to `0x1`, we can pass it from the config.

birchmd:
### Description
Closes #465 

### Changes
- Pass chain ID from genesis config into the EVM
- Construct proper `SessionId` type during transaction simulation

### Testing
- New test `test_block_env_evm_contract`


server/src/tests/evm_contracts/block_env.rs
```diff
@@ -0,0 +1,201 @@
...
+        // 2.b check `msg.sender` as a view call
+        let request = view_contract(contract_address, getMsgSenderCall::SELECTOR.to_vec());
+        let sender = request.from.unwrap();
+        let response = ctx
+            .eth_call(request, BlockNumberOrTag::Latest)
+            .await
+            .unwrap();
+        assert_eq!(
+            response,
+            [[0_u8; 12].as_slice(), sender.0.as_slice()].concat()
```
bulenttastan, line 77: nit: This is EVM to MoveVM address conversion, could be done differently but that's OK.

bulenttastan (approved):
Looks and works great.

app/src/query.rs
```diff
@@ -248,10 +249,22 @@ impl<'app, D: Dependencies<'app>> ApplicationReader<'app, D> {
...
+                timestamp: block.0.header.timestamp,
+                prev_randao: block.0.header.mix_hash,
+                chain_id: self.genesis_config.chain_id,
+            })
+            .unwrap_or_default();
         call_transaction(
             transaction,
             &self.state_queries.resolver_at(height)?,
             &self.evm_storage,
+            block_header,
```
ilchu, line 267: As a side note, why do we add it for `eth_call` only, and not the other view calls? I don't quite see the convenience of this.

ilchu (approved):

evm-ext/src/native_impl.rs
```diff
@@ -259,48 +269,57 @@ fn evm_error(e: EVMError<DbError>) -> SafeNativeError {
...
+        tx_type: 0,
+        kind: transact_to,
+        value,
+        data: data.into(),
+        // Nonce and chain id can be ignored because replay attacks
+        // are prevented at the MoveVM level. I.e. replay will
+        // never occur because the MoveVM will not accept a duplicate
+        // transaction
+        nonce: 0,
+        chain_id: None,
```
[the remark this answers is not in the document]
birchmd: Yes, these two instances of `chain_id` are different. Here is the chain ID that is part of the EVM transaction. In the other place it is setting the chain ID the EVM itself expects. The EVM will reject a transaction that has the wrong chain ID. The purpose of this check is to prevent replaying transactions across chains without the user's permission. As noted in the comment, we don't bother setting the chain ID in the EVM transaction (and therefore skip the chain ID check in the EVM) because the chain ID has already been checked at the MoveVM level, so we do not need to check it again.

app/src/query.rs
birchmd: The header is only required for view calls that trigger execution and there are only two of those: `eth_call` and `eth_estimateGas`. The latter was already correctly using the header in its execution because it follows the same flow as for transactions. `eth_call` needed fixing because it uses a flow separate from transactions which was missing the header.

PR #469, opened 2025-07-25
feat(blockchain): Add ABI to Move module response in `StateQueries`

code-dogg:
### Description
Adds the ABI response structure, as seen in [Aptos API spec](https://fullnode.mainnet.aptoslabs.com/v1/spec#/operations/get_account_module).

### Changes
- Declare Move module ABI structure taken from Aptos.
- Add ABI into Move module response in `StateQueries`.

### Testing
Using clippy and CI.

### Notes
#415


blockchain/src/state/read/response/abi.rs
```diff
@@ -0,0 +1,160 @@
+use {
+    move_core_types::{ability::AbilitySet, account_address::AccountAddress},
+    std::fmt,
+};
+
+/// A Move module
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct MoveModule {
```
bulenttastan, line 8: Are these types available on `aptos-core` repo to pull in directly?

blockchain/src/state/read/response/conversions.rs
```diff
@@ -0,0 +1,298 @@
...
+            StructFieldInformation::Native => (true, vec![]),
+            StructFieldInformation::Declared(fields) => (
+                false,
+                fields
+                    .iter()
+                    .map(|f| self.new_move_struct_field(f))
+                    .collect(),
+            ),
+            StructFieldInformation::DeclaredVariants(..) => {
+                // TODO(#13806): implement for enums. Currently we pretend they don't have fields
```
birchmd, line 206: This todo is referencing an aptos issue. Is there a way we can avoid inheriting their tech debt?

birchmd (commented):
I agree with @bulenttastan that it would be nice if we did not need to duplicate definitions which exist in the aptos code base.

ilchu (commented):
Another thing we could do here is directly use Aptos's API maybe? There's a `aptos-api-types` crate and their `MoveModuleBytecode` [type](https://github.com/aptos-labs/aptos-core/blob/ee9d74edde1b6bad787dade8718f74e32e6af2d5/api/types/src/move_types.rs#L1207) seems to be exactly what we need here.

Though of coursethere could be some discrepancies between this and e.g. Sui when we start supporting it properly, and so we might need a tailor-made solution eventually.

PR #470, opened 2025-07-25
fix(docker): build independent services first

ilchu:
### Description

Deploy action in CI started failing after the most recent update to docker stack due to broken paths. That happened because the actual image used there was the one being pulled from the registry instead of the one built locally. This PR introduces a fix to that behavior and some rolling update tagging for peace of mind.

### Changes

- Staged builds
- New `:previous` tag

### Testing
Temporary added PR trigger to the action, will remove if successful.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
 
 concurrency:
   group: ${{ github.workflow }}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -22,16 +24,48 @@ jobs:
...
+          done
         continue-on-error: true
-      - name: Build images
-        run: docker compose build
-      - name: Push images
-        run: docker compose push
+
+      # Stage 1: Build and push base images first
+      - name: Build base images
+        run: |
```
bulenttastan, line 53: We don't need the pipe `|`, just have the run:
```
run: docker compose build optimism op-move geth
```

bulenttastan (approved):

PR #478, opened 2025-08-04
refactor(api): make engine API closer to the spec

### LINKED ISSUE #63: Implement ForkchoiceUpdate according to the Ethereum specification

ilchu:
### Description

Make our engine API closer to the spec by running more validations. Some tests are still failing, and I'll probably write at least some new ones, but otherwise good to go.

Fixes #63.

### Changes

- `forkchoiceUpdated` and `newPayload` with more thorough validation
- Some fresh fields from OP spec
- Changed the way mempool txs are treated. Instead of being drained unconditionally, they're only cleaned up at the end of block build now

### Testing

- [x] Old tests passing: new_payload API test had a sneaky hard-coded value in it, so I refactored it with more sensible flow. 
- [x] Write some new tests: integration-style one for the `no_tx_pool` flag.

### DIFF
--- a/app/src/command.rs
+++ b/app/src/command.rs
@@ -4,7 +4,11 @@ use {
         actor::UnrecoverableAppFailure,
         input::{WithExecutionOutcome, WithPayloadAttributes},
     },
-    alloy::{consensus::Receipt, primitives::Bloom, rlp::Encodable},
+    alloy::{
+        consensus::{Receipt, constants::EMPTY_WITHDRAWALS},
+        primitives::{B256, Bloom},
+        rlp::Encodable,
+    },
     umi_blockchain::{
         block::{BaseGasFee, Block, BlockHash, BlockRepository, ExtendedBlock, Header},
         payload::{PayloadId, PayloadQueries},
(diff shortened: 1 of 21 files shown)


birchmd (changes requested):
Nice work overall! Just one issue with the mempool draining.

app/src/command.rs
```diff
@@ -191,7 +215,25 @@ impl<'app, D: Dependencies<'app>> Application<'app, D> {
...
-        in_progress_payloads.finish_id(block, transactions.into_iter().map(Into::into));
+
+        // Mempool should only be cleaned after all other building stages succeeded
+        if !no_tx_pool {
+            // Payload transactions always come first in the block
+            new_transactions
+                .iter()
+                .skip(attributes_txs_len)
+                .for_each(|tx| {
+                    self.mem_pool.remove_by_hash(tx.tx_hash(), tx.signer());
```
birchmd, line 226: I just noticed that it would be more efficient to remove from the mempool by signer + nonce instead of signer  + hash because then we do not need to iterate through all the transactions from that sender looking for the right hash. We have access to the nonce here so I think it makes more sense to do that.

ilchu: Good catch! I should have considered that from the start, now that I think about it. Only had to add some failure conditions to tx casting, but it's even more explicit that way.

birchmd (approved):

PR #480, opened 2025-08-06
feat(api): Implement `web3_clientVersion`

### LINKED ISSUE #474: Implement `web3_clientVersion` API
OpenZeppelin upgradable proxy contract calls this API for some reason. We need this so deployments can succeed.

https://ethereum.org/en/developers/docs/apis/json-rpc/#web3_clientversion

Client version is composed of `binary_name/build_version/os_architecture/rust_version`

The values can be static or dynamically generated. Static values can be:
- Binary name: op-move
- Build version: v0.1.1-devnet-stable
- OS arch: linux-x86_64
- Rust version: rust1.85.1

ilchu:
### Description

Implements a client version endpoint with help of build scripts as e.g. target is only exposed at that stage, and rust version itself does not have a variable preset for it. 

Fixes #474.

### Changes

- New method
- New build script for the app crate

### Testing

Due to the generated environment varaibles being runtime-dependent, the new endpoint unit tests are mostly about the general shape of the response, not the content.

### DIFF
--- a/api/src/methods/client_version.rs
+++ b/api/src/methods/client_version.rs
@@ -35,7 +35,6 @@ mod tests {
             .unwrap()
             .split("/")
             .collect::<Vec<_>>();
-        dbg!(&parts);
 
         assert_eq!(parts.len(), 4);
 
--- a/app/src/query.rs
+++ b/app/src/query.rs
@@ -41,8 +41,9 @@ impl<'app, D: Dependencies<'app>> ApplicationReader<'app, D> {
 
     pub fn client_version(&self) -> String {
         format!(
-            "op-move/{}/{}/{}",
+            "op-move/v{}-{}/{}/{}",
             env!("CARGO_PKG_VERSION"),
+            env!("GIT_HEAD"),
             env!("TARGET_TRIPLET"),
             env!("RUSTC_VERSION")
         )
--- a/app/build.rs
+++ b/app/build.rs
@@ -12,7 +12,19 @@ fn main() {
         .nth(1)
         .unwrap_or("unknown_version")
         .to_string();
-    println!("cargo::rustc-env=RUSTC_VERSION=rust_{}", version);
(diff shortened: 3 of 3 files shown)


birchmd (approved):
Nice job!

api/src/methods/client_version.rs
```diff
@@ -0,0 +1,70 @@
...
+        });
+
+        let actual_response = execute(request, &reader).await.unwrap();
+
+        let parts = actual_response
+            .as_str()
+            .unwrap()
+            .split("/")
+            .collect::<Vec<_>>();
+        dbg!(&parts);
```
bulenttastan, line 38: nit: we probably don't need this.

app/src/query.rs
```diff
@@ -39,6 +39,15 @@ impl<'app, D: Dependencies<'app>> ApplicationReader<'app, D> {
         self.genesis_config.chain_id
     }
 
+    pub fn client_version(&self) -> String {
+        format!(
+            "op-move/{}/{}/{}",
```
bulenttastan, line 44: This is what op-move vs OP Mainnet returns:
```
op-move/0.1.0/aarch64-apple-darwin/rust_1.85.1
Geth/v1.101503.4-stable-2b9abb39/linux-amd64/go1.24.2
```
| | name | version | target | compiler |
| -- | -- | -- | -- | -- |
| **Umi** | op-move | 0.1.0 | aarch64-apple-darwin | rust_1.85.1 |
| **OP** | Geth | v1.101503.4-stable-2b9abb39 | linux-amd64 | go1.24.2 |

We don't need to fix them cause I don't know how useful this RPC is, but here are the differences I notice:
1. The version needs a `v` prefixed, the rest is OK
2. Target triplet has the `OS` first then the `arch`
3. No underscore needed after `rust`

bulenttastan (approved):
Looks good. This is the response:
`op-move/v0.1.0-2ce07db11/aarch64-apple-darwin/rust1.85.1`

PR #481, opened 2025-08-06
Feat(api): mv_listModules and mv_listResources

birchmd:
### Description
Adds new APIs for listing the modules and resources available on an account.

### Changes
- Skip List data structure implementation
- Use skip list to index modules and resources on each account (this is included in the trie)
- New APIs: `mv_listModules` and `mv_listResources`

### Testing
- New tests on skip list implementation
- New integration tests: `test_mv_list_modules` and `test_mv_list_resources`


state/src/skip_list.rs
```diff
@@ -0,0 +1,771 @@
...
+        trie.insert(insert_key.key_hash().as_slice(), &end_of_list)?;
+
+        return Ok(());
+    };
+
+    if element < first_value.as_ref() {
+        // The new element needs to become the head of the list.
+
+        // Randomly pick the height for the old head (now second element).
+        let insert_level = (pick_insert_height(rng) - 1) as u32;
```
ilchu, line 119: The `-1`'s purpose here evades me a bit. And also a bit of a nit:

```suggestion
        let reinsert_level = (pick_insert_height(rng) - 1) as u32;
```

as it's hard to keep in mind that it's meant for the *old* first node, not the new one.

state/src/skip_list.rs
```diff
@@ -0,0 +1,771 @@
...
+
+            let fv_value = fv_key.trie_value(trie)?;
+            match fv_value {
+                Some(value) => {
+                    trie.remove(fv_key.key_hash().as_slice())?;
+                    trie.insert(insert_key.key_hash().as_slice(), &value.serialize())?;
+                }
+                None => trie.insert(insert_key.key_hash().as_slice(), &end_of_list)?,
+            }
+        }
```
ilchu, line 156: So we completely replace the head node on levels higher than its picked (re)insertion level, and prepend it with the new node on all lower ones, right? I believe this snippet would benefit from having common parts taken out of the matches, like insertion of the new first node and old first key/value construction.

state/src/skip_list.rs
```diff
@@ -0,0 +1,771 @@
...
+        };
+        current_key.level = lower_level;
+        continue;
+    }
+}
+
+fn pick_insert_height<R: Rng>(rng: &mut R) -> usize {
+    let mut result = 1;
+    let d = Bernoulli::new(0.5).expect("Probability is valid.");
+    while d.sample(rng) {
```
ilchu, line 530: Wouldn't we want to bound it from above here? Otherwise there's a small but non-negligible chance we could end up with too many redundant layers, and we never do any excess level pruning.

state/src/skip_list.rs
```diff
@@ -0,0 +1,771 @@
...
+        let Some(lower_level) = current_key.level.checked_sub(1) else {
+            return Ok(prevs);
+        };
+        current_key.level = lower_level;
+        continue;
+    }
+}
+
+fn pick_insert_height<R: Rng>(rng: &mut R) -> usize {
+    let mut result = 1;
```
ilchu, line 528: Why is the picked height is by default at least 1, though the base level is 0-indexed?

state/src/skip_list.rs
```diff
@@ -0,0 +1,771 @@
...
+        trie: &EthTrie<D>,
+    ) -> Result<SkipListHeadValue<'static, T>, TrieError> {
+        let trie_key = self.key_hash();
+        SkipListHeadValue::read_trie(&trie_key, trie)
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct SkipListHeadValue<'a, T: Clone> {
+    pub max_levels: u32,
```
ilchu, line 260:
```suggestion
    pub current_highest_level: u32,
```

ilchu (commented):
It's very neat how the index is constructed on the fly and serialized, but I wonder if there's going to be performance implications for accounts with long resource / module lists, as on every insert we're doing quite a number of old value read and writes and it could quickly amplify.

Is there absolutely no way we could do some interning of values? E.g. one way I'm thinking it could be possible is by storing whole serialized levels of the list under a single key. As all the search / insertion / deletion operations work by the level anyway, this should not present a problem, and BCS supports sequences. Or would the constant re-encoding burden be higher than the trie hopping, what do you think?

state/src/skip_list.rs
```diff
@@ -0,0 +1,771 @@
...
+    fn kind() -> u8 {
+        0
+    }
+}
+
+impl Listable for Identifier {
+    fn kind() -> u8 {
+        1
+    }
+}
```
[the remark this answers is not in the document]
birchmd: Yes, I agree an enum would be simpler, but I think the trait fits better with the type generics used in defining the skip list functions. It does also have the nice side effect that the skip list tests can use `u64` even though the production code does not.

state/src/skip_list.rs
```diff
@@ -0,0 +1,771 @@
...
+
+type SkipListPair<T> = (SkipListKey<'static, T>, SkipListValue<'static, T>);
+
+fn search_item<T, D>(
+    account: AccountAddress,
+    max_levels: u32,
+    first_value: Cow<'static, T>,
+    element: &T,
+    trie: &EthTrie<D>,
+) -> Result<Vec<Option<SkipListPair<T>>>, TrieError>
```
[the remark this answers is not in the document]
birchmd: Renamed in https://github.com/UmiNetwork/op-move/pull/481/commits/3db5679c165b54b7fccbdfcaefe2103275227b98

state/src/skip_list.rs
birchmd: I tried to clarify things in https://github.com/UmiNetwork/op-move/pull/481/commits/4d9bd0bd0844ce610a8bba77b35f3c3f298c69fc

server/src/tests/listing_apis.rs
```diff
@@ -0,0 +1,286 @@
...
+
+async fn get_all_modules(ctx: &TestContext<'static>, address: AccountAddress) -> Vec<Identifier> {
+    let address = address.to_eth_address();
+    let mut all_modules = Vec::new();
+    loop {
+        let mut modules = ctx
+            .mv_list_modules(address, all_modules.last(), None, BlockNumberOrTag::Latest)
+            .await
+            .unwrap();
+        if modules.is_empty() {
```
bulenttastan, line 243: nit: Probably not needed optimization for test, less code is more readable I guess.

ilchu (approved):

state/src/skip_list.rs
```diff
@@ -0,0 +1,800 @@
...
+            SkipListIterator::<u64, MemoryDB>::new_from_start(AccountAddress::ZERO, &3, &trie, 0)
+                .unwrap();
+        let values: Vec<u64> = iter.map(|v| v.unwrap()).collect();
+        assert_eq!(values, vec![5, 8]);
+    }
+
+    #[test]
+    fn test_delete() {
+        let mut trie = mock_list();
+        delete_item::<u64, MemoryDB>(AccountAddress::ZERO, &5, &mut trie).unwrap();
```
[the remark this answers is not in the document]
birchmd: Good catch! I fixed the handling of deleting the first element of the list in https://github.com/UmiNetwork/op-move/pull/481/commits/0d6b9b38754cf5751759cd0d71b9513ce9bb4c0f

server/src/tests/listing_apis.rs
```diff
@@ -0,0 +1,286 @@
...
+        // Listing API handles the case where there are no resources
+        let signer = PrivateKeySigner::from_slice(&[0xaa; 32]).unwrap();
+        let address = signer.address();
+        let account_resources = ctx
+            .mv_list_resources(address, None, None, BlockNumberOrTag::Latest)
+            .await
+            .unwrap();
+        assert_eq!(account_resources, Vec::new());
+
+        // When a new resources is created the listing API sees it
```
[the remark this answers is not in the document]
birchmd: Fixed in https://github.com/UmiNetwork/op-move/pull/481/commits/f7e1c1ac876e224034ff89ecf0f4517bb8334f8c

state/src/skip_list.rs
```diff
@@ -0,0 +1,913 @@
...
+    }
+
+    Ok(())
+}
+
+/// Each type that implements `Listable` must have a unique value for `kind`.
+/// It is used as a discriminator to make the byte-level representation of the keys
+/// in the trie unique.
+pub trait Listable {
+    fn kind() -> u8;
```
bulenttastan, line 286: Is this used part of the serialization in the `element`? There won't be any overlap between modules and resources, right, when inserted in the list?

bulenttastan: Yes, checked and the key hashes indeed change with different `kind`s, so that differentiates the types.

bulenttastan (approved):

PR #482, opened 2025-08-11
Feat(api): Get storage at API for EVM

bulenttastan:
### Description
Implement `eth_getStorageAt` API for EVM only. It returns `0x0` if the account or the storage is empty.

### Changes
- Add storage access methods in query and model
- Expose the storage with RPC API

### Testing
✓ Integration test with the storage of the counter contract.

### DIFF
--- a/server/src/tests/test_context.rs
+++ b/server/src/tests/test_context.rs
@@ -191,8 +191,7 @@ impl TestContext<'static> {
             "method": "eth_getStorageAt",
             "params": [address, index, "latest"]
         });
-        let value_hex: String = handle_request(request, &self.queue, self.reader.clone()).await?;
-        let value = U256::from_str_radix(value_hex.trim_start_matches("0x"), 16)?;
+        let value: U256 = handle_request(request, &self.queue, self.reader.clone()).await?;
         Ok(value)
     }
 


blockchain/src/state/read/model.rs
```diff
@@ -162,6 +162,29 @@ pub trait StateQueries {
...
+        &self,
+        evm_storage: &impl StorageTrieRepository,
+        account: Address,
+        index: U256,
+        height: BlockHeight,
+    ) -> Result<U256, state::Error> {
+        let resolver = self.resolver_at(height)?;
+
+        // Read account info to get the storage root
+        let evm_db = ResolverBackedDB::new(evm_storage, &resolver, &(), 0);
```
birchmd, line 177: Nit: could use `height` here instead of `0`. It doesn't matter because the block height is not relevant to the lookup we are doing on the EVM state here, but it might be nice to pass it in properly anyway.

birchmd (approved):
Nice job!

It would probably also be good to add some usage of this API in the tests in `server/src/tests/evm_contracts/`, but that can be done as a separate PR.

birchmd (approved):

PR #483, opened 2025-08-11
feat: Add Aptos API with get account module endpoint

code-dogg:
### Description
Adds an endpoint of the Aptos API as documented here: https://aptos.dev/rest-api/operations/get_account_module

### Changes
- Add a new `warp` route that leads to [`get_account_module`](https://aptos.dev/rest-api/operations/get_account_module) endpoint

### Testing
Using clippy and CI

### Notes
#415


birchmd (commented):
I left a comment about REST vs JSON RPC. We should all be on the same page about how we're presenting these query methods.

Either way though, I think it would be nice to have integration tests similar to those included in #481

server/src/lib.rs
```diff
@@ -174,6 +180,19 @@ fn serve(
     let content_type =
         HeaderMap::from_iter([(CONTENT_TYPE, HeaderValue::from_static("application/json"))]);
 
+    let aptos_route = warp::path!("v1" / "accounts" / AccountAddress / "module" / String)
```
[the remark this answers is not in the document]
bulenttastan: Yes correct, we should handle the calls via JSON RPC.

We will probably add a translation REST API layer in the future, but even if we do, we don't need to handle them manually like this. We would use something like `poem-openapi` to define them in Open API.

api/src/methods/get_module.rs
```diff
@@ -0,0 +1,234 @@
+use {
+    crate::{json_utils::parse_params_3, jsonrpc::JsonRpcError},
+    umi_app::{ApplicationReader, Dependencies},
+    umi_shared::primitives::{Address, ToMoveAddress},
+};
+
+pub async fn execute<'reader>(
```
bulenttastan, line 7: Suggestion: We should add comments with example request params, since this is a non-standard RPC.

api/src/methods/get_module.rs
```diff
@@ -0,0 +1,234 @@
...
+    request: serde_json::Value,
+    app: &ApplicationReader<'reader, impl Dependencies<'reader>>,
+) -> Result<serde_json::Value, JsonRpcError> {
+    let (number, account, module): (_, Address, String) = parse_params_3(request)?;
```
birchmd, line 11: Bikeshedding: I think it is strange to have the block number come first. In Ethereum RPC functions it always comes last.

PR #487, opened 2025-08-14
Fix(genesis): Test root hash to match genesis config

bulenttastan:
### Description
Root hash is changed in [genesis config](https://github.com/MovedNetwork/op-move/blob/99532c8c1d6a18beba12389f71d0600c1fffd38f/genesis/src/config.rs#L98) but not in the default server config.

### Changes
- Update the root hash in default config with the one in genesis config

### Testing
Integration test works, it was failing before.

### Notes
Why don't we use `GenesisConfig::default()` here? Instead of duplicating and potentially missing to update both values.


birchmd (approved):
> Why don't we use `GenesisConfig::default()` here? Instead of duplicating and potentially missing to update both values.

I agree this would be a good change.

PR #488, opened 2025-08-14
Feat(genesis): Compute OP stack expected state root

### LINKED ISSUE #225: Generate state root hash for genesis block internally

birchmd:
### Description
Closes #225 

Hooray for `alloy` having a one-liner for this!

### Changes
- Compute genesis state root using L2 config.

### Testing
- Checked against previously hard-coded state root.

### DIFF
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -1,6 +1,6 @@
 use {
     crate::mirror::MirrorLog,
-    alloy::consensus::{EMPTY_OMMER_ROOT_HASH, EMPTY_ROOT_HASH},
+    alloy::consensus::{proofs::state_root_unhashed, EMPTY_OMMER_ROOT_HASH, EMPTY_ROOT_HASH},
     jsonwebtoken::{DecodingKey, Validation},
     move_core_types::account_address::AccountAddress,
     std::{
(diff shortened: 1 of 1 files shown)


server/src/lib.rs
```diff
@@ -311,9 +311,7 @@ fn create_genesis_block(
         parent_beacon_block_root: Some(B256::ZERO),
         parent_hash: B256::ZERO,
         receipts_root: EMPTY_ROOT_HASH,
-        state_root: B256::new(hex!(
-            "30b67e4b5ef34eacb9e083c07fd5578982c2cb4e0ee1dc0a14d72b99a28ed80e"
-        )),
+        state_root: state_root_unhashed(genesis_config.l2_contract_genesis.alloc.clone()),
```
bulenttastan, line 314: Wow, that's amazing 🚀
While you're at it, can you pull the default genesis config here?

birchmd: Done in https://github.com/UmiNetwork/op-move/pull/488/commits/4bcb135b94fa981efdc44c1aca012a5e153e144c

bulenttastan (approved):

PR #492, opened 2025-08-19
Chore: refactor gas price type as u128

### LINKED ISSUE #227: Unify gas price `uint` type
Identify what's the correct `uint` type on Ethereum and make the conversion in the beginning to keep using the same type.
This is causing to use `saturating_to()` calls for uint type casting.

https://github.com/MovedNetwork/op-move/blob/5065d42637a91b01a04a2c4608e547a58e71d266/moved/src/state_actor.rs#L660

### LINKED ISSUE #255: Add invariant error for gas price `u256` to `u64` casting
Either consider an assertion with invariant error or remove the `// TODO` comment.

https://github.com/MovedNetwork/op-move/blob/5bff3017103056dbc9c1b2d4a300fc22089fd81e/moved/src/types/session_id.rs#L84

birchmd:
### Description
Closes #227 
Closes #255 

This PR sets the gas price type based on alloy (u128 for general gas prices, u64 for base fee). Aptos only accepts u64 gas prices, so there is also a check at the start of executing canonical transaction that the gas price fits in a u64. It is an invalid transaction error if it does not fix because the priority fee is set as part of the transaction and the base fee is already a u64. Note this should never be a limitation in practice because a u64 variable can represent 18 ETH, and the smallest transaction costs 21k gas, which means if the gas price exceeds a u64 then even a basic transfer would cost millions of dollars; quite an unrealistic scenario.

### Changes
- Represent general gas price as u128
- Represent base fee as u64.

### Testing
- Existing tests
- New `test_invalid_gas_price`

### DIFF
--- a/blockchain/src/block/gas.rs
+++ b/blockchain/src/block/gas.rs
@@ -1,6 +1,6 @@
 //! This module is concerned about calculating fees charged for gas usage.
 
-use {std::cmp::Ordering, umi_shared::primitives::U256};
+use std::cmp::Ordering;
 
 /// Determines amount of fees charged per gas used in transaction execution.
 ///
@@ -14,8 +14,8 @@ pub trait BaseGasFee {
         &self,
         parent_gas_limit: u64,
         parent_gas_used: u64,
-        parent_base_fee_per_gas: U256,
-    ) -> U256;
+        parent_base_fee_per_gas: u64,
+    ) -> u64;
 }
 
 /// Calculates base fee per gas according to the Ethereum model based on EIP-1559.
@@ -49,7 +49,7 @@ pub struct Eip1559GasFee {
     /// * The greater the value the smaller the increase or decrease of the base fee per gas.
     /// * This value has to be greater than zero.
     /// * A value of 1 makes the greatest fee increases or decreases.
-    base_fee_max_change_denominator: U256,
+    base_fee_max_change_denominator: u128,
 }
 
 impl Eip1559GasFee {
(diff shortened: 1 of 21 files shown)


blockchain/src/block/gas.rs
```diff
@@ -76,30 +76,33 @@ impl BaseGasFee for Eip1559GasFee {
...
+                    .saturating_mul(u128::from(gas_target - parent_gas_used))
+                    / u128::from(gas_target)
                     / self.base_fee_max_change_denominator;
 
                 parent_base_fee_per_gas.saturating_sub(delta)
             }
             Ordering::Equal => parent_base_fee_per_gas,
-        }
+        };
+        result.try_into().unwrap_or(u64::MAX)
```
code-dogg, line 105: Should we return the newly introduced error instead of capping it at `u64::MAX`?

code-dogg (approved):
We should probably add a test for the newly introduced error case at some point.

blockchain/src/block/gas.rs
birchmd: Yes, good call. Done in https://github.com/UmiNetwork/op-move/pull/492/commits/873fe7f23578eca531701111ff454984990c1113 . I also added a new test in that commit.

birchmd: Actually that was a different spot that I changed. I misunderstood what part of the code this comment was on. Here I think it makes sense to cap it because it is computing a base fee mechanically based on an algorithm (it has nothing to do with user transactions). So this line is handling a possible overflow by saturating the result, which is correct behaviour (though this case should never come up).

PR #501, opened 2025-08-27
feat(api): support engine API v4

ilchu:
### Description

First step toward OP bump support.

### Changes

- Most changes to `newPayload` processing
- `getPayload` just falls back to V3 like `op-geth`
- feature gated one format check that interferes with old logic


birchmd (approved):
LGTM

Seems a little silly to have a new field that cannot be used, but if that's how OP stack is handling it then that is what we must do as well.

PR #502, opened 2025-08-27
Feat(API): mv_getTableItem

### LINKED ISSUE #497: Implement `mv_getTableItem` to retrieve table data with a handle id
An account resource for a table will return the table's handle id. We need an RPC method to retrieve the actual table data from the Tables native extension.

RPC endpoint will look similar to [GetTableItem](https://fullnode.mainnet.aptoslabs.com/v1/spec#/operations/get_table_item) API.

birchmd:
### Description
Closes #497 

### Changes
- New logic for parsing JSON value into Move value (following Aptos standard)
- New state queries method for getting table items
- New API method `mv_getTableItem`

### Testing
- New integration test `test_mv_get_table_item`

### Notes
- We can't use the JSON conversion logic in `aptos-api-types` because it is intertwined with the Aptos-specific trait `StateView`. Including the logic in our own code is not that bad since it is not much code.


api/src/methods/get_table_item.rs
```diff
@@ -0,0 +1,32 @@
+use {
+    crate::{json_utils::parse_params_3, jsonrpc::JsonRpcError},
+    move_table_extension::TableHandle,
+    umi_app::{ApplicationReader, Dependencies},
+};
+
+/// Fetches `resource` by name that belongs to the `account` in the blockchain state corresponding
```
bulenttastan, line 7: It's copied from `get_resource`, we should adapt it to table.

blockchain/src/state/read/model.rs
```diff
@@ -200,15 +206,43 @@ pub trait StateQueries {
...
+
+        let value_type = request
+            .value_type
+            .try_into()
+            .map_err(|_| UserError::IncorrectTypeLayout)?;
+
+        let resolver = self.resolver_at(height)?;
+        let converter = UmiMoveConverter::new(&resolver);
+
+        let key = converter.try_into_vm_value(&key_type, request.key)?;
```
bulenttastan, line 234: The key has to be sent in as a string for `u64`. Is that expected?
The test example is for `Table<bool, u64>` and sending key as a bool works in the request, but for `Table<u64, u64>` it expects the key to be a string:
```
"key_type": "u64",
"value_type": "u64",
"key": "2"
```

blockchain/src/state/read/convert.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub fn try_into_vm_value_struct(
+        &self,
+        layout: &MoveStructLayout,
+        value: serde_json::Value,
+    ) -> Result<MoveValue> {
+        let (struct_tag, field_layouts) =
+            if let MoveStructLayout::WithTypes { type_, fields } = layout {
+                (type_, fields)
+            } else {
+                return Err(incorrect_type_layout(()).into());
```
ilchu, line 131: Is a wrapping function for the error here just serving brevity? I feel it would be more straightforward to just use the variant itself.

server/src/tests/table_api.rs
```diff
@@ -0,0 +1,153 @@
...
+    let signature = signer.sign_transaction_sync(&mut tx).unwrap();
+    let signed_tx = TxEnvelope::Eip1559(tx.into_signed(signature));
+    ctx.execute_transaction(signed_tx).await.unwrap()
+}
+
+fn create_move_tables_contract_bytecode(address: Address) -> Vec<u8> {
+    let bytecode_hex = std::fs::read_to_string("src/tests/res/tables.hex").unwrap();
+    let bytecode = hex::decode(bytecode_hex.trim()).unwrap();
+    crate::tests::integration::set_module_address(bytecode, address)
+}
```
ilchu, line 153: Can we have some short description detailing actual contract code? E.g. what functions it has and such, or a link from where it was taken if it's Aptos standard something.

ilchu (approved):
Just some clarification points.

blockchain/src/state/read/convert.rs
birchmd: I was doing it for consistency with the other error handling (`map_err` on results). But I agree it is less clear than just using the variant directly. Fixed in https://github.com/UmiNetwork/op-move/pull/502/commits/1ea77e5332e749fa461ce3a91bbcb402f65592af

PR #507, opened 2025-09-08
feat(engine): pass `L2ToL1MessagePasser` storage root in block header

ilchu:
### Description

Implements [this part](https://specs.optimism.io/protocol/isthmus/exec-engine.html#l2tol1messagepasser-storage-root-in-header) of Optimism L2 execution engine spec, setting the withdrawals root in header to the value of the storage root of `L2toL1MessagePasser` contract for ease of L2 output root generation for historic blocks without access to an archival node.

### Changes

- Withdrawals root is set in every block post-exec 
- It is also set in genesis appropriately


app/src/command.rs
```diff
@@ -31,6 +31,10 @@ use {
     umi_state::State,
 };
 
+#[cfg(feature = "op-upgrade")]
+const L2_TO_L1_MESSAGE_PASSER_ADDRESS: alloy::primitives::Address =
```
birchmd, line 35: You also have this constant defined in `app/src/lib.rs`, it could be imported from there instead of duplicated.

birchmd (approved):
Some minor nits, but otherwise looks good. It would also be nice if there was a test for this change. But maybe it makes more sense to add it when we finish the upgrade and remove the feature flag.

birchmd (approved):

PR #509, opened 2025-09-11
fix(engine): harden payload handling

### LINKED ISSUE #508: Fix expired payload timestamps
Investigating logs from a stalling dev instance has shown that prior to a stall op-move sends an empty block without a single transaction back to op-node. This means that it received some deposits-only attributes via a `forkchoiceUpdate` which contained only previously seen txs and filtered them out from the resulting block. 

Besides being incorrect behaviour, it also indicates further issues that were the cause of this occurrence - namely, the op-node previously noticing our block header had a timestamp mismatch by 1 second with the corresponding payload attributes. This triggered a reorg and made op-node send that payload again, but we didn't process it properly.

The solution for this, besides load testing that's being done by @birchmd to determine the cause of initial slowdown that led to this, is to bring our logic closer to what `op-geth` would do, again. The points that I've noticed so far that might be contributing to the issue:

1. Our `forkchoiceUpdate` handling doesn't look through payload IDs upon receiving the attributes, but unconditionally sends a block build command. Instead we should just return a valid status as a response immediately in that case.
2. We don't check for mismatch with parent block timestamp (i.e. if the difference is bigger than block time). Neither do we timeout old build jobs or have any other safeguards. While we don't want to go all the way of e.g. making block build async, e.g. checking system time could be beneficial here.

I might find some more issues upon a closer examination and I'll be fixing those accordingly.

ilchu:
### Description

Some minute changes to the way payload attributes are handled around timestamps and in progress jobs.

Fixes #508.

### Changes

- Ensure monotonic timestamp relative to parent
- Avoid sending block build commands for existing payloads
- Remove the respective payload from in progress map if the block build fails

### Notes

When going through our main loop again, I noticed two very itchy issues that we never addressed, so I'll try to pick it up next:
* The actor is never automatically restarted on a failed built, but left hanging indefinitely, whereas we could have a supervisor of sorts to do just that
* There's no transactionality to the changes done during block build: e.g. if some DB access fails on receipt write, the Move/EVM changes would have already been committed and there's no rollback or anything, leaving the state inconsistent. A more correct approach would be to batch all changes together and roll back on any failure

### DIFF
--- a/app/src/command.rs
+++ b/app/src/command.rs
@@ -52,6 +52,7 @@ impl<'app, D: Dependencies<'app>> Application<'app, D> {
             return Ok(());
         }
         let in_progress_payloads = self.payload_queries.get_in_progress();
+        // If there is a job with this id already present, nothing to do
         if in_progress_payloads.start_id(id).is_err() {
             return Ok(());
         }
@@ -60,9 +61,9 @@ impl<'app, D: Dependencies<'app>> Application<'app, D> {
         let no_tx_pool = attributes.no_tx_pool.unwrap_or(false);
 
         let transactions = if no_tx_pool {
-            // Though it should be a rare event without follower nodes, `op-node` can demand
-            // to rebuild a block deterministically from payload attributes only. We log it
-            // as a rare event for diagnostic purposes.
+            // Though it is a relatively rare event without follower nodes, `op-node` can demand
+            // to rebuild a block deterministically from payload attributes only. This also happens
+            // when an L2 reorg is triggered. We log it for diagnostic purposes.
             tracing::warn!(
                 "Building from payload attributes only, with no mempool txs: {attributes:?}"
             );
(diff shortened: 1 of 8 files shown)


app/src/command.rs
```diff
@@ -100,6 +101,17 @@ impl<'app, D: Dependencies<'app>> Application<'app, D> {
                 UnrecoverableAppFailure
             })?
             .expect("Block repository is non-empty (must always at least contain genesis)");
+        // TODO: also check block time (preferrably dynamically from a config) and
```
birchmd, line 104: It's probably worth making a new issue for this. You can include the details about how op-geth pulls txs as a background task to make sure the block time is never exceeded while building.

birchmd (approved):
Nice job! I like that this is a relatively small fix; most of the changed lines are in tests.

PR #512, opened 2025-09-17
fix(engine): limit concurrency of non-engine requests

ilchu:
### Description

Limits the maximal number of in-progress http requests for the server via a tokio semaphore.

### Changes

- New component in the server filter that's only used by the http path

### Notes
Hard-coded values for the concurrency are suboptimal, but we can include configurability in a future PR.


birchmd:
Closing in favour of #515

PR #519, opened 2025-09-26
build(docker): Remove `builder` stage and minimize layers in `optimism` stage

code-dogg:
### Description

Reducing a number of stages and layers reduces the time and space needed for the docker image build.

### Changes
- Remove `builder` stage
- Minimize layers in `optimism` stage

### Testing
```
docker compose build
```

### DIFF
--- a/docker/Dockerfile.optimism
+++ b/docker/Dockerfile.optimism
@@ -1,58 +1,54 @@
 # First build all the dependencies
-FROM golang:1.22-alpine AS builder
+FROM golang:1.22-alpine AS optimism
 
-# Install build dependencies
-RUN apk add --no-cache git make bash wget
-
-# Set working directory
-WORKDIR /build
-
-# Clone and build optimism at specific commit
-RUN git clone https://github.com/ethereum-optimism/optimism . && \
-  git checkout f2e5a7a5 && \
-  make cannon-prestate op-node op-batcher op-proposer
-
-# Download and extract foundry binaries
+# Declare argument for target architecture supplied automatically by the build system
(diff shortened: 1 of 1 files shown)


ilchu (changes requested):
I'm not completely sure this kind of change either decreases the build times or space requirements as claimed. If we're speaking disk space usage during build, both forge and make commands probably generate some intermediate artifacts, and being more granular about copying only the executables could have some small savings.

As for build time, layers should help caching intermediate results. Surely, we're not doing it with our own code, and always checking out a specific commit so far, but in the best case scenario some of the steps still get skipped on repeat builds (though because we run a git clone, it surely gets funky inside as docker seems to struggle with web interactions within the image).

Even in case my suppositions are wrong and nothing gets cached between layers, both pre- and post-patch docker images should be pretty equivalent. Have you run any rough comparisons that have shown some improvement with this PR?

code-dogg:
@ilchu The point is the stage called `build` is removed. The cache set to `max` mode stores every stage in the multi-stage docker file. This makes it one less stage to keep.

We use `max` mode, because without it, the pre-final stages need to be rerun. 

https://docs.docker.com/build/cache/backends/#cache-mode

> In min cache mode (the default), only layers that are exported into the resulting image are cached, while in max cache mode, all layers are cached, even those of intermediate steps.

> While min cache is typically smaller (which speeds up import/export times, and reduces storage costs), max cache is more likely to get more cache hits. Depending on the complexity and location of your build, you should experiment with both parameters to find the results that work best for you.

PR #520, opened 2025-09-26
build(docker): Remove local directory cache

code-dogg:
### Description

The `docker-compose.yml` has a `local` cache. This is a cache that uses a local directory.

This causes that the cache gets exported. Since it is local only, the export and import is not needed, and takes time to generate and over 10 gigabytes of storage to keep.

### Changes
- Remove local directory cache from `docker-compose.yml`

### Testing
```
docker compose build
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,9 +18,6 @@ server/*.log
 # Docker mounted volumes
 docker/volumes
 
-# Docker buildx cache directory
-.buildx-cache
-
 # Optimism cloned repository
 server/src/tests/optimism
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,10 +5,6 @@ services:
       context: ./
       dockerfile: ./docker/Dockerfile.optimism
       target: op-node
-      cache_from:
-        - type=local,src=.buildx-cache/op-node
-      cache_to:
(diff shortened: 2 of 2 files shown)


ilchu (approved):
We could have discussed it when you approved #506. I do not have a hard stance on this, as ultimately it's a convenience issue, and the import/export also happen purely locally. So it's a trade off between repeat builds and some disk space, and I would lean towards speed of iteration being more important. Still, if you deem it better this way, either way's fine by me.

code-dogg:
@ilchu There is no tradeoff. Let me explain.

The `cache_from` and `cache_to` directives are used to import and export caches to remote sources. But the cache is always kept inside the docker host on the machine that is building.

The `local` cache is only useful if the cache directory is shared amongst multiple docker hosts.

The `local` cache gives no caching improvements for local building.

You don't lose cache with this.

https://docs.docker.com/build/cache/backends/local/

> The local cache store is a simple cache option that stores your cache as files in a directory on your filesystem, using an [OCI image layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md) for the underlying directory structure. Local cache is a good choice if you're just testing, or if you want the flexibility to self-manage a shared storage solution.

PR #523, opened 2025-09-29
build(docker): Initialize geth during image build

code-dogg:
### Description
Runs the geth initialization during image build which includes:
* Initialize keystore for deterministic account addresses
* Prefund dev accounts
* Deploy Optimism factory deployer

It also brings:
* Graceful shutdown: When you stop the geth container, it actually propagates the SIGINT to the process and waits for it to properly finish.
* Restartability: Thanks to the graceful shutdown and #521 the container can be stopped and started again and will pickup exactly where it was stopped.
* Speed of docker operations: The geth container shutdown speed is now decreased from 10s to a couple of milliseconds. Mostly due to the fact that the interrupt signal reaches geth this time.

### Changes
- Launch geth during image build and run initialization transactions
- Implement graceful shutdown for the geth container

### Testing
```
docker compose build geth && docker compose up geth
```

### DIFF
--- a/docker/Dockerfile.geth
+++ b/docker/Dockerfile.geth
@@ -56,14 +56,38 @@ RUN \
 COPY --from=build /volume/build/bin/geth /usr/local/bin/geth
 COPY --from=foundry /cast /usr/local/bin/cast
 COPY --from=foundry /forge /usr/local/bin/forge
-COPY docker/wait-for-it.sh /usr/local/bin/wait-for-it
-COPY docker/prefund.sh prefund.sh
-COPY .env .env
-COPY docker/geth.sh entrypoint.sh
-COPY docker/keystore.sh keystore.sh
-COPY docker/geth-init.sh geth-init.sh
+COPY --chmod=0777 docker/wait-for-it.sh /usr/local/bin/wait-for-it
+COPY --chmod=0777 docker/prefund.sh prefund.sh
+COPY --chmod=0777 .env .env
+COPY --chmod=0777 docker/geth.sh entrypoint.sh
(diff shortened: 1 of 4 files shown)


docker/Dockerfile.geth
```diff
@@ -57,14 +57,38 @@ RUN \
...
+COPY --chmod=0777 docker/geth-init.sh geth-init.sh
 
-# Set entrypoint to geth binary
-RUN chmod +x entrypoint.sh /usr/local/bin/wait-for-it prefund.sh keystore.sh geth-init.sh
+RUN \
+  # Create datadir for geth
+  mkdir -p ./l1_datadir \
+  # Initialize keystore in the datadir
+  && ./keystore.sh \
+  # Launch geth node to receive transactions
```
ilchu, line 72: It seems like a more precise thing to say here is that we run this command to pre-init the keystore and datadir, and then run the process again from those starting points.

code-dogg: Not sure how to integrate your feedback.

> ... we run this command to pre-init the keystore and datadir ...

For that we run the commands above:

```
  # Create datadir for geth
  mkdir -p ./l1_datadir \
  # Initialize keystore in the datadir
  && ./keystore.sh \
```

> ... and then run the process again from those starting points.

I don't quite understand.

The geth node is launched to receive transactions sent by running `geth-init.sh` that prefund dev accounts and deploy Optimism factory deployer as explained here:

```
  # Send transactions that prefund accounts and deploy Optimism factory deployer
  && ./geth-init.sh \
```

birchmd (approved):

PR #531, opened 2025-10-01
ci: Cache Rust build artifacts

code-dogg:
### Description
Adds [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache) action to cache cargo and target using [Github cache API](https://cli.github.com/manual/gh_cache).

### Changes
- Add cache step before all steps that produce Rust build artifacts

### Testing
Using Github actions

### Notes
The action was sometimes failing with `exit code 143`. Upon some research, it was discovered that this code is when a process receives SIGTERM, which, in the context of Github actions, probably means that it was interrupted due to exhaustion of resources, likely memory.

I added a different ELF linker `mold`, which seems to mitigate this issue while providing a significant compilation speed boost (about 2-3 minutes).

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -19,12 +19,18 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Install mold to speed up ELF linking and decrease memory usage
+        uses: rui314/setup-mold@v1
       - name: Free up disk space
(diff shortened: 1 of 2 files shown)


ilchu (commented):
I guess the main source of those spurious interrupts is just the huge size of cache, as this action caches all intermediate artifacts by default in `target`. E.g. my current size of that folder is 85G. It could be turned off, but that would also kill most of the benefit as then it will only be caching the cargo registry.

PR #532, opened 2025-10-01
feat(server): expose prometheus endpoint

ilchu:
### Description

Exposes a scraping endpoint for metrics and instruments RPC request handling.

### Changes

- Handler installation in main
- Metrics calls in API for RPC and in app for block build

### Testing

Just `cargo check`. Will try to run it on a local stack later.

### Notes

An alternative route was to make the endpoint available via warp, but this change was more minimal (as e.g. that would need different headers).

Besides the work here, I'll need to setup the so-called [Op Agent](https://cloud.google.com/monitoring/agent/ops-agent/prometheus) in GCP and restart the devnet, and hopefully we'll be able to see the new metrics in the [dashboard](https://console.cloud.google.com/monitoring/metrics-explorer;duration=P7D?hl=en&project=blockchain-451501) immediately after.

### DIFF
--- a/api/src/request.rs
+++ b/api/src/request.rs
@@ -4,6 +4,7 @@ use {
         jsonrpc::{JsonRpcError, JsonRpcResponse},
         method_name::MethodName,
     },
+    metrics::{counter, gauge, histogram},
     umi_app::{ApplicationReader, CommandQueue, Dependencies},
     umi_blockchain::payload::NewPayloadId,
 };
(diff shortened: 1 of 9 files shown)


api/src/request.rs
```diff
@@ -93,7 +94,13 @@ where
         return Err(JsonRpcError::missing_method(request));
     }
 
-    match method {
+    let method_str = format!("{method:?}");
```
birchmd, line 97: Instead of using the debug format of the `MethodName` enum, maybe it would be better to use the string value from the JSON directly. The JSON method name is prescribed by the Ethereum RPC spec, while our internal representation of it may change in the future. So think the metric label will be more stable if the JSON value is used.

Something like

```rust
let method_str = json_utils::get_field(&request, "method")
    .as_str()
    .ok_or(JsonRpcError::missing_method(request.clone()))?;

let method: MethodName = method_str.parse()?;
```

birchmd (approved):

PR #535, opened 2025-10-02
feat(api): timestamp incoming requests via an nginx proxy

ilchu:
### Description

For purposes of potential traffic replay or otherwise incident analysis, having the actual timestamps of incoming requests that are produced externally from op-move would be beneficial. This PR tries to achieve that via installing a minimalistic proxy up front that forwards all traffic but appends the receive time.

With this the appended value should seamlessly appear in logs as we capture the whole request body as is.

### Changes

- A small conf file for nginx
- A new service in the docker compose

### Testing

Local docker build

### DIFF
--- a/api/src/request.rs
+++ b/api/src/request.rs
@@ -22,7 +22,7 @@ pub enum SerializationKind {
 pub struct RequestModifiers<'a, A, P> {
     is_allowed: A,
     payload_id: &'a P,
-    request_start_ms: u128,
+    request_start_ms: Option<u128>,
     serialization_tag: SerializationKind,
 }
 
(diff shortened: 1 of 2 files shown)


nginx/rpc.conf
```diff
@@ -0,0 +1,17 @@
+server {
+  listen 8545;
+
+  # stamp request start time (milliseconds since epoch)
+  set $req_start_ms $msec;
+  proxy_set_header X-Req-Start-Ms $req_start_ms;
```
birchmd, line 6: As discussed on the call, this approach of adding a timestamp to the header makes sense, but we also need to do something with the heading inside `op-move`. Maybe building off the metrics PR we could expose a "real" request duration metric which is based on the header timestamp. Then a difference between the `op-move` measured duration and this "real" duration would be a measure of congestion on the tokio side.

We also discussed the possibility of using this header timestamp to perform exact local replays of devnet runs, but that can be left as future work.

ilchu: Added the new metrics in fae2936 and created #543 for further work.

PR #563, opened 2025-10-24
Feat(forkchoice): Allow building forks

birchmd:
### Description
Allows op-move to build a block tree instead of just a block chain. This is useful when a block build requested by `op-node` times out so `op-node` tries to build a new block with that same parent.

This is a pretty large PR, but most of the changes are related to test code (lots of tests were using height as a primary key for state queries and this no longer works when there can be multiple blocks at the same height). The key business logic changes are reflected in changes to trait definitions and the implementations of those traits.

### Changes
- Include forkchoice state in the persisted database
- Reference state roots by block hash instead of height
- Parent used for block build is based on the forkchoice
- Active state root (used in block building) is set during forkchoice updates
- Better validation in `engine_forkchoiceUpdated`

### Testing
- Existing tests
- New integration test `test_forks`

### DIFF
--- a/api/src/methods/new_payload.rs
+++ b/api/src/methods/new_payload.rs
@@ -63,7 +63,7 @@ async fn inner_execute<'reader>(
     let withdrawals_root = {
         use umi_blockchain::state::StateQueries;
 
-        match app.state_queries.evm_storage_root_at(
+        match app.state_queries.evm_storage_root_from_block_hash(
             &app.evm_storage,
             umi_app::L2_TO_L1_MESSAGE_PASSER_ADDRESS,
             execution_payload.block_hash,
--- a/blockchain/src/state/read/model.rs
+++ b/blockchain/src/state/read/model.rs
@@ -269,9 +269,9 @@ pub trait StateQueries {
         Ok(storage.get(&index)?.unwrap_or_default())
     }
 
-    /// Queries the blockchain state version corresponding with block `height` for the value of an
+    /// Queries the blockchain state version corresponding with block `hash` for the value of an
     ///  EVM `account` storage root. Returns 0x0..00 if no account was found.
-    fn evm_storage_root_at(
+    fn evm_storage_root_from_block_hash(
         &self,
         evm_storage: &impl StorageTrieRepository,
         account: Address,


app/src/command.rs
```diff
@@ -285,19 +301,85 @@ impl<'app, D: Dependencies<'app>> Application<'app, D> {
...
-        Ok(())
+        Ok(Some(block_hash))
     }
 
     pub fn add_transaction(&mut self, tx: NormalizedEthTransaction) {
         self.mem_pool.insert(tx);
     }
 
+    #[tracing::instrument(level = "debug", skip(self))]
+    pub(crate) fn forkchoice_update(
```
ilchu: Nit: we could have a docstring here too, if there's a new one on block build already.

api/src/methods/new_payload.rs
```diff
@@ -384,15 +384,27 @@ mod tests {
         let head_hash = B256::new(hex!(
             "781f09c5b7629a7ca30668e440ea40557f01461ad6f105b371f61ff5824b2449"
         ));
-        let genesis_block = Block::default()
-            .into_extended_with_hash(head_hash)
-            .with_value(U256::ZERO);
+        let genesis_block = {
+            let mut tmp = Block::default();
+            tmp.header.state_root = genesis_config.initial_state_root;
```
ilchu: Is it done for the same reason as what the comment says in `app::command::forkchoice_state()`, i.e. to propagate full move state to the app?

blockchain/src/block/read.rs
```diff
@@ -17,14 +20,58 @@ pub trait BlockQueries: Debug {
...
+    fn get_forkchoice_state(&self, storage: &Self::Storage) -> Result<ForkchoiceState, Self::Err>;
+
+    fn height_to_hash(&self, storage: &Self::Storage, height: u64) -> Result<B256, Self::Err>;
+
+    /// This function returns `true` if the given block hash `maybe_ancestor` is reachable
+    /// via parent hash links in the blockchain starting from the block hash `head` or
+    /// `maybe_ancestor == head`.
+    /// Note: the default implementation is rather inefficient (it always walks the whole chain
+    /// from `head` to `maybe_ancestor`); implementors of this trait should provide a more streamlined
+    /// implementation (e.g. using a skip list).
```
ilchu: I've looked up how (op) geth does it, and apparently instead of walking the chain they try to do a cleaner separation of side and canonical chains. This means they always look up the head block in FCU by hash and height from header, and a mismatch signals that the canonical chain has changed, which triggers a reorg. IDK if that would be a better design but still.

app/src/query.rs
```diff
@@ -402,61 +417,95 @@ impl<'app, D: Dependencies<'app>> ApplicationReader<'app, D> {
...
-            limit,
-        )?)
+        let hash = self.resolve_height(height)?.hash;
+        Ok(self
+            .state_queries
+            .move_list_resources(address.to_move_address(), hash, after, limit)?)
     }
 
-    fn resolve_height(&self, height: BlockNumberOrTag) -> Result<u64> {
+    fn resolve_height(&self, height: BlockNumberOrTag) -> Result<alloy::rpc::types::eth::Header> {
```
ilchu: Nit:
```suggestion
    fn resolve_height_to_header(&self, height: BlockNumberOrTag) -> Result<alloy::rpc::types::eth::Header> {
```
or something along these lines should sound clearer as I wouldn't expect getting a header back from just a height/tag as the most default thing.

storage/rocksdb/src/block.rs
```diff
@@ -114,24 +124,19 @@ impl BlockQueries for RocksDbBlockQueries {
...
 
-    fn latest(&self, db: &Self::Storage) -> Result<Option<u64>, Self::Err> {
-        Ok(db
-            .iterator_cf(&height_cf(db), IteratorMode::End)
-            .next()
-            .transpose()?
-            .map(|(height, _)| u64::from_key(height.as_ref())))
+    fn height_to_hash(&self, db: &Self::Storage, height: u64) -> Result<B256, Self::Err> {
+        let maybe_hash = db
+            .get_pinned_cf(&fc_cf(db), height.to_key())?
```
ilchu: Shouldn't we be querying height CF here? And AFAICT forkchoice is keyed by a single string const anyways.

server/src/dependency/rocksdb.rs
```diff
@@ -209,10 +211,20 @@ impl<'db> umi_app::Dependencies<'db> for RocksDbReaderDependencies {
...
+                    tracing::error!("on_payload callback failed: {e:?}");
+                    Error::DatabaseState
+                })?;
+            state
+                .state_queries
+                .push_state_root(hash, state.state.state_root())
+                .map_err(|e| {
+                    tracing::error!("on_payload callback failed: {e:?}");
+                    Error::DatabaseState
+                })
```
ilchu: I believe this could interfere with withdrawals root setting after OP upgrade. Previously the state root was updated right after executing the transactions, and before the header was constructed or the storage root of `L2ToL1MessagePasser` queried. But the `on_payload()` callback is applied much later down the line, so the state root at the time of setting might be stale.

Another thing I'm not sure about here is whether resolver needs to be updated too, considering that withdrawals root calculation relies on the EVM DB, virtually making it equivalent to the `state::switch_state_root()` call as is done in `command::forkchoice_update()` itself.

ilchu (changes requested):
It definitely feels nicer to have the execution API spec followed more closely, and the complexity added is not too overwhelming. Also the drive-by test refactors are always welcome. Among my comments only two last ones are semantic and almost certain to require changes, everything else is more cosmetic or just nits.

storage/rocksdb/src/block.rs
birchmd: Fixed in https://github.com/UmiNetwork/op-move/pull/563/commits/9b73e68560267e8f0bdbf4ba256b86b4739d7124

blockchain/src/block/in_memory.rs
```diff
@@ -1,20 +1,37 @@
 use {
     crate::{block::ExtendedBlock, payload::PayloadId},
+    alloy::rpc::types::engine::ForkchoiceState,
```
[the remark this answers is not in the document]
birchmd: Fixed in https://github.com/UmiNetwork/op-move/pull/563/commits/0fcda6872bdfda6543633ebde6f17107020faaa8

api/src/methods/mod.rs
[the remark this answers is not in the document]
birchmd: I removed the unnecessary clippy annotation in https://github.com/UmiNetwork/op-move/pull/563/commits/d5fd657a107d1d672f7fe5ba22881741595e2558

I agree these functions are now close and could maybe be unified. But I will leave it for now because I don't think there is too much value in continuing with test refactors immediately.

ilchu (approved):
Just some minor cosmetics, but otherwise very much LGTM. Let's see if the rest of the team have any additions.

storage/rocksdb/src/block.rs
```diff
@@ -1,19 +1,21 @@
...
-        block::{BlockQueries, BlockRepository, BlockResponse, ExtendedBlock},
+        block::{BlockQueries, BlockRepository, BlockResponse, ExtendedBlock, ForkchoiceState},
         transaction::ExtendedTransaction,
     },
     umi_shared::primitives::B256,
 };
 
 pub const BLOCK_COLUMN_FAMILY: &str = "block";
 pub const HEIGHT_COLUMN_FAMILY: &str = "height";
+pub const FC_COLUMN_FAMILY: &str = "forkchoice";
```
bulenttastan: Is height database still needed? (the line above)

birchmd: Yes, the height column family is still needed because we still keep a height to hash mapping for when queries come with explicit heights (as opposed to a tag like `latest`). But this is now the only case where height is used because height is no longer the primary key for state lookup; height maps to a block hash (the mapping is forkchoice dependent) and the block hash is the primary key for looking up a state root.

api/src/methods/call.rs
```diff
@@ -109,8 +110,18 @@ mod tests {
                     },
                 payload_id: U64::from(i),
             };
-              state_channel.send(msg).await.unwrap();
-             state_channel.reserve_many(10).await.unwrap();
+            state_channel.send(msg).await.unwrap();
```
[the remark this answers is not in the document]
birchmd: Added a helper function in https://github.com/UmiNetwork/op-move/pull/563/commits/595a5cc3dfca357da2d911aa66fd3ea3b2fa81e2

PR #567, opened 2025-11-03
feat(op-upgrade): fjord L1 gas fee

ilchu:
### Description

The Fjord hard fork of OP stack (that literally happened a few weeks after our pinned Optimism version) introduced a different [algorithm](https://specs.optimism.io/protocol/fjord/exec-engine.html#fjord-l1-cost-fee-changes-fastlz-estimator) for L1 fee calculation that's supposed to represent final sizes more accurately. This PR introduces that logic under the upgrade feature flag.

This is also a prerequisite for the `miner_setMaxDASize` API call that calculates the DA size according to this spec.

### Changes

- L1 gas input has a new field that's based on FastLZ-compressed calldata size
- Updated gas structures used under the upgrade flag
- Tests fixed (most unrelated to this PR, could have been broken under the recent test refactor) 

### Testing

Old tests pass both with `--features op-upgrade` and without.

### Notes
When revising the L1 fee logic, I was a bit confused why we don't scale it back by `1e6` like OP does even under Ecotone, but I had a feeling we might have discussed it a long time ago. For the new algo I kept the same scale (i.e. no division by a million) nonetheless.

### DIFF
--- a/execution/src/gas.rs
+++ b/execution/src/gas.rs
@@ -270,20 +270,19 @@ impl FjordGasFee {
 #[cfg(feature = "op-upgrade")]
 impl L1GasFee for FjordGasFee {
     fn l1_fee(&self, input: L1GasFeeInput) -> U256 {
-        use umi_shared::primitives::I256;
-
         // The spec <https://specs.optimism.io/protocol/fjord/exec-engine.html#fjord-l1-cost-fee-changes-fastlz-estimator>
-        // returns a `U256` as the final result, so we can widen the types in advance while
-        // adding the intercept sign (due to clunkiness of converting from non-ruint types).
-        let intercept = -I256::from_raw(U256::from(Self::INTERCEPT_ABS));
+        // returns a `U256` as the final result, so we can widen the types in advance.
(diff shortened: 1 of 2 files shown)


execution/src/gas.rs
```diff
@@ -205,6 +235,83 @@ impl L1GasFee for EcotoneGasFee {
...
+        let estimated_size_scaled = {
+            let scaled = I256::from(fast_lz_coef * fast_lz_size) + intercept;
+            let min_scaled = I256::from(min_tx_size * U256::from(1_000_000));
+            U256::from(scaled.max(min_scaled))
+        };
+
+        let weighted_gas_price = Self::GAS_PRICE_MULTIPLIER * self.base_fee_scalar * self.base_fee
+            + self.blob_base_fee_scalar * self.blob_base_fee;
+
+        // We scale down by 1e6 instead of 1e12 to preserve the previous Ecotone omission of
```
birchmd: I am not sure why we skipped the 1e6 factor originally. This is something we will maybe need to review again in the future.

PR #568, opened 2025-11-03
build(docker): Add a replica instance of `op-node` and `op-move`

### LINKED ISSUE #303: Explore multi-node setup

code-dogg:
Closes #303 

### Description
Adds a p2p connected, fully independent, replicated `op-node` and `op-move` service that uses the main services to sync.

The publicly exposes services are still just the main ones.

### Changes
- Add `op-move-rep` to `docker-compose.yml`
- Add `op-node-rep` to `docker-compose.yml`

### Testing
Using docker

### DIFF
--- a/app/src/command.rs
+++ b/app/src/command.rs
@@ -183,7 +183,7 @@ impl<'app, D: Dependencies<'app>> Application<'app, D> {
             Some(storage_root)
         };
         #[cfg(not(feature = "op-upgrade"))]
-        // Has to be `keccak256(rlp(empty_string_code))`, so we can reuse the ommers value
+        // Has to be `keccak256(rlp(empty_string_code))`
         let withdrawals_root = Some(alloy::consensus::constants::EMPTY_WITHDRAWALS);
 
         let header = Header {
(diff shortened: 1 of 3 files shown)


app/src/command.rs
```diff
@@ -184,7 +184,7 @@ impl<'app, D: Dependencies<'app>> Application<'app, D> {
         };
         #[cfg(not(feature = "op-upgrade"))]
         // Has to be `keccak256(rlp(empty_string_code))`, so we can reuse the ommers value
```
bulenttastan: We can remove the comment part `, so we can reuse...`

bulenttastan (approved):
Tested the replica locally, works great 👏

app/src/command.rs
```diff
@@ -184,7 +184,7 @@ impl<'app, D: Dependencies<'app>> Application<'app, D> {
...
-        let withdrawals_root = Some(alloy::consensus::constants::EMPTY_OMMER_ROOT_HASH);
+        let withdrawals_root = Some(alloy::consensus::constants::EMPTY_WITHDRAWALS);
```
birchmd: Thanks for fixing this!

